PR #13, opened 2022-11-11
Print improvements

FreyJo:


nurkanovic (approved):
New print log is cleaner!

PR #14, opened 2022-11-11
Add lambda00 to fix implicit Euler, Cleanup parameters

FreyJo:
TODO: fix for Step.


nurkanovic (approved):
Your changes are fine, the naming is convient. 
I believe that for "Step" lambda_00 has the dimension 2*n_theta. 
We should use n_lambda in the next update, which is 
for Stewart n_lambda  = n_theta and for Step n_lambda = 2*n_theta.

PR #15, opened 2022-11-14
Fix lambda00 for multi simplex model and x0 in homotopy_solver

FreyJo:
#14  was not meant to be merged already. Sorry should have clearified this.
I found two related bugs and now the friction block example works again.


nurkanovic (approved):
This looks good! Thanks. 
I am only wondering why do not we have also for the "step" case a function for lambda_00, but only an expression, cf. line 52 in homotopy_solver.
We might consider to make a casadi function for it as well.

PR #16, opened 2022-11-17
Add basic testing using gitub actions, and an example matlab test.

apozharski:
Simple sanity check test example (that doesn't necessarily check for correctness currently, just a lack of matlab warnings).


FreyJo (approved):
Good job, looks great! :+1:

PR #19, opened 2022-11-23
Fix bugs with use_fesd = 0 and refactor everything around time transformations

nurkanovic:


FreyJo (approved):
Nice, even more tests! :+1:

PR #21, opened 2022-11-27
Polished step equilibration, polished all examples, more compact solver outputs. 

nurkanovic:


FreyJo (approved):
Look good to me.
I think you can remove the outdated script `step_equilibration_constrains` now.

PR #26, opened 2022-12-15
Index refactor

apozharski:
Refactoring index definitions. ~Very work in progress.~


src/matlab/create_nlp_nosnoc.m
```diff
@@ -140,13 +140,33 @@
...
+problem.ind_h = [];
+problem.ind_elastic = [];
+problem.ind_sot = []; % index for speed of time variable
+problem.ind_boundary = []; % index of bundary value lambda and mu
+problem.ind_t_final = []; % Time-optimal problems: define auxilairy variable for the final time.
+% Constraints
+problem.g = {};
+problem.lbg = [];
+problem.ubg = [];
+problem.ind_g_clock_state = [];
```
FreyJo, line 166: I just checked out of curiosity, `ind_g_clock_state` seems to never be used.
Can be removed later of course.

src/matlab/create_nlp_nosnoc.m
```diff
@@ -512,95 +456,83 @@
...
+                end
+                % Add contribution to the end state, Attention qj changes with time scaling!
+                Xk_end = Xk_end + D(j+1)*X_ki_stages{j};
+                % Add contribution to quadrature function
+                J = J + B(j+1)*qj*h_ki;
+                % Append IRK equations to NLP constraint
+                problem = add_constraint(problem, h_ki*fj - xp, zeros(n_x, 1), zeros(n_x, 1));
+
+              case 'differential'
+                % Evaluate Differential and Algebraic Equations at stage points
```
FreyJo, line 488: There seem to be some inconsistent indents 2 vs 4 spaces.

FreyJo (approved):
2 minor comments.
LGTM

nurkanovic (commented):
a short description for add_variable (e.g., what are possible values for type)
 would be useful, even though it is almost apparent what needs to be done.

nurkanovic (commented):
To make create_nlp_nosnoc shorter, I suggest to make some small stupid functions, e.g.,
create_empty_problem for initializing the problem struct

nurkanovic (commented):
we often use (which was my initial mistake) cells for stacking functions and variables, g = {problem.g{:}, g}, and do vertcat(g(:)) at the end. we could immediately just do 
problem.g = [problem.g, g], right @FreyJo ?

FreyJo:
> we often use (which was my initial mistake) cells for stacking functions and variables, g = {problem.g{:}, g}, and do vertcat(g(:)) at the end. we could immediately just do problem.g = [problem.g, g], right @FreyJo ?

Right, it should be possible to just do `vertcat` in the new helper functions.
But, as Anton said this is one cleanup step and there are surely more to come.

FreyJo:
> a short description for add_variable (e.g., what are possible values for type) would be useful, even though it is almost apparent what needs to be done.

Personally, I would not put comments in such a self explaining function, which is anyway just a helper function for developers, and not part of the front-end.

PR #30, opened 2023-01-06
Classification

apozharski:
- created class structure for nosnoc formulation roughly in line with `nosnoc_py`.
- x0 is now a parameter.
- added support for global/time varying parameters.


src/matlab/create_nlp_nosnoc.m
```diff
@@ -214,786 +83,69 @@
...
-model.ind_boundary = problem.ind_boundary;
-model.n_cross_comp = n_cross_comp;
-model.h = h;
-model.h_k = h_k;
-model.p_val = p_val;
-model.n_cross_comp_total = sum(n_cross_comp(:));
+model.p_val = problem.p0;
+
+model.structured_ind.x = problem.ind_x;
+model.structured_ind.x0 = problem.ind_x0;
```
FreyJo, line 140: why are the indices copied to model?
IMO `model` should be agnostic of `problem` stuff.
`problem` and these indices are very specific to discretization etc.

src/matlab/homotopy_solver.m
```diff
@@ -39,34 +39,37 @@
...
-    lambda00 = full(model.lambda00_fun(x0));
+    lambda00 = full(model.lambda00_fun(x0, model.p_global_val));
 elseif strcmp(settings.pss_mode, 'Step')
 %     c_x = full(model.c_fun(x0));
 %     lambda00 = [ max(c_x, 0); min(c_x, 0)];
-    lambda00 = full(model.lambda00_fun(x0));
+    lambda00 = full(model.lambda00_fun(x0, model.p_global_val));
 end
-p_val = [model.p_val(:); lambda00(:)];
+p_val = [model.p_val(:);x0(:); lambda00(:)];
```
FreyJo, line 64: In Python, the parameters are organized as:
```
p_val = np.concatenate((prob.model.p_val_ctrl_stages.flatten(), np.array([sigma_k, tau_val]), lambda00, x0))
```
i.e. model parameters (given by user), homotopy parameters (introduced when setting up the problem for the homotopy solver), `lambda00, x0` - "eliminated" variables.
It seems that `model.p_val`, gets augmented with the homotopy parameters, which is not ideal IMO, but no "merge blocker" :)

src/matlab/homotopy_solver.m
```diff
@@ -39,34 +39,37 @@
...
 
 % TODO remove try!
 try
     complementarity_stats = [full(comp_res(w0, p_val))];
 catch
     w0 = w0(1:length(model.w));
+    w0
+    p_val
```
FreyJo, line 72: These kind of `try`s seem really sketchy.
If you found why its needed maybe add a comment.
But you didnt introduce it of course.

src/matlab/homotopy_solver.m
```diff
@@ -86,8 +89,7 @@
     fprintf('\niter\tsigma\t\tcompl_res\tCPU time\tNLP iters\tstatus\n')
 end
 
-
-while (complementarity_iter) > comp_tol && ii < N_homotopy && sigma_k > sigma_N
+while (complementarity_iter) > comp_tol && ii < N_homotopy && (sigma_k > sigma_N || ii == 0)
```
FreyJo, line 92: why is the ii == 0 part needed?

src/matlab/homotopy_solver.m
FreyJo: Nervermind, I think `model.p_val` only contains the homotopy parameters and you just didnt touch that.

src/matlab/polishing_homotopy_solution.m
```diff
@@ -29,7 +29,9 @@
 end
 %% unfold
 import casadi.*
+settings_bkp = settings;
 unfold_struct(settings,'caller');
+settings = settings_bkp;
```
FreyJo, line 34: This seems weird

src/matlab/model_reformulation_nosnoc.m
```diff
@@ -36,7 +36,9 @@
 end
 
 unfold_struct(model,'caller');
+settings_bkp = settings;
 unfold_struct(settings,'caller')
+settings = settings_bkp;
```
FreyJo, line 41: Why is that needed?

FreyJo (commented):
This is more a superficial review, but it is really hard to do more given the size of this PR.
For the future, I suggest to not add features within a refactoring PR.
I think the added feature here is support for model parameters, not sure if there are more.
Maybe you can also add a short summary of what changed in the PR description.

src/matlab/model_reformulation_nosnoc.m
apozharski: Because for a reason I still have not hunted down, `unfold_struct` sometimes clobbers some members of its argument in the caller scope.  I will drop a comment on these.

src/matlab/homotopy_solver.m
apozharski: for the direct reformulation  `sigma_k` is set to sigma_N directly. This ensures that even if the user does this that one homotopy iteration will run regardless (which is the intended outcome).

src/matlab/ControlStage.m
```diff
@@ -0,0 +1,218 @@
+classdef ControlStage < NosnocFormulationObject
```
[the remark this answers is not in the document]
apozharski: In python cross complementarities are only done at the `FiniteElement` level. In order to cleanly separate what happens on each level I thought it useful to have a `ControlStage` object that handles things like defining stage variables and control stage level cross complementarities.

src/matlab/model_reformulation_nosnoc.m
FreyJo: ok, sorry I actually saw a comment at some point. It indeed seems very weird behavior though :sweat_smile:

src/matlab/ControlStage.m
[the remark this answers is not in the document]
FreyJo: Ok, I don't have a strong opinion.
I just see that now there are all index vectors, like `ind_theta` on `FiniteElement`, `ControlStage` and `Problem` level.
Which I am not sure if it simplifies things or not.

src/matlab/IrkRepresentation.m
```diff
@@ -0,0 +1,7 @@
+classdef IrkRepresentation
+    enumeration
+        integral
```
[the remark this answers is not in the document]
apozharski: Yep this will be done in a second pass that also changes all the examples. These currently match the strings used for these settings.

examples/matlab/time_freezing_hystheresis/time_optimal_control_car/main_car_hysteresis_ocp.m
```diff
@@ -46,15 +46,19 @@
...
+settings.initial_lambda = 0.0;
+settings.initial_theta = 0.0;
+settings.initial_mu = 0.0;
+
 %% Model Settings
 model.fuel_cost_on = 0;
 model.N_finite_elements = 3;
 model.N_stages = 10;
-model.T = 5;
+model.T = 1;
```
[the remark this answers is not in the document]
nurkanovic: This one is not unnecessary. One can still provide a `T` in a time optimal problem as well.
This is the length of the "numerical time" and it can be different than 1. 
Mathematically, there optimal time is `T_opt = T*s_sot` . There is some ambiguity of course. 
You can always set `T = 1` (or any number you like) and let `s_sot` do the rest of the job.
However, in practice putting some higher/different `model.T`  can improve the convergence, as you iterate less over `s_sot` and the problem is quite nonlinear in these directions.

src/matlab/homotopy_solver.m
nurkanovic: I think I might have an explanation. 
At some point not all functions were defined w.r.t. to the correct size of `p_val` as  `p` was modified at several places (e.g., after the introduction of `lambda_00`).
However, comp_res did not depend initially on any `p_val` initially but was defined to depend on it, but with wrong size.

Of course, it should work without a `try` and `catch`. 
Should be an issue anymore, I believe?

src/matlab/create_nlp_nosnoc.m
apozharski: Yep this probably should not live in model.

nurkanovic (approved):
I could not look everything line by line, but I have looked into details in several places.
More importantly, I checked all examples and played with different settings and the outcomes are reasonable. The code does what it should do.
I have a few smaller remarks, which do not change how the code works, but rather simplify some things. However, they are more suitable for our to-do lists and for future PRs.

Thanks for the good work in this huge PR! 🚀🥳

PR #31, opened 2023-01-26
Simple General Inclusions

apozharski:
Adding expert user functionality to manually define `alpha` and `f_x` to handle non-fillipov systems. 

Also added printing of objective.

### DIFF
--- a/src/matlab/homotopy_solver.m
+++ b/src/matlab/homotopy_solver.m
@@ -79,7 +79,7 @@ complementarity_iter = 1;
 ii = 0;
 
 if print_level >= 3
-    fprintf('\niter\tsigma\t\tcompl_res\tobjective\tCPU time\tNLP iters\tstatus\n')
+    fprintf('\niter\t\tsigma\t\tcompl_res\tobjective\tCPU time\tNLP iters\tstatus\n')
 end
 
 while (complementarity_iter) > comp_tol && ii < N_homotopy && (sigma_k > sigma_N || ii == 0)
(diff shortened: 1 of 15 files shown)


nurkanovic (commented):
LGTM, except some small verbose things. 
the alignment is a bit disturbed in the homotopy solver, at least in my command window.
When I copy it here it looks nice for some reason.
```

iter		sigma		compl_res	objective	CPU time	NLP iters	status
1		1.00e+00	2.10e+02	59.899		7.083		496		Solved_To_Acceptable_Level
```
quite  confusing...

Anyway, it seems to me that the homotopy solver still prints the $\ell_1$ cross comp and not $\ell_{\infty}$
In the final verbose after the problem is solve i would not print the $\ell_1$ anymore but just the $\ell_\infty$.

src/matlab/homotopy_solver.m
```diff
@@ -63,15 +63,8 @@
...
-% TODO remove try!
-try
-    complementarity_stats = [full(comp_res(w0, p_val))];
-catch
-    w0 = w0(1:length(model.w));
-    w0
-    p_val
-    complementarity_stats = [full(comp_res(w0, p_val))];
-end
+complementarity_stats = [full(comp_res(w0, p_val))];
```
nurkanovic, line 66: comp_res still computes $\ell_1$ without the cross terms, right? 
nice that the stupid try-catch is resolved.

apozharski: Yep, it still does, I haven't quite finished that, but hopefully will today in a follow on PR. As for the formatting I made it look nice in my command window so it seems somehow our machines are handling tabs differently. I'm going to look if matlab has a nice table output formatter, and if it does switch to using that.

nurkanovic: Lets get the current PR through as the inclusion part is anyway good and let the verbose be part of the next one.

nurkanovic (approved):

PR #32, opened 2023-01-28
User algebraics

apozharski:


examples/matlab/gene_regulation/irma_model.m
```diff
@@ -52,10 +52,29 @@
...
+        else
+            beta = SX.sym('beta', 2);
+            model.f_alg = beta - [alpha(1)*(1-alpha(7)); alpha(2)*(1-alpha(5))];% lifted equations
+            s = [1, alpha(6);
+             1, beta(1);
+             1, alpha(3);
+             alpha(2), beta(2);
+             1, alpha(4)];
+        end
+        model.z_user = beta;
```
nurkanovic, line 72: I was thinking to name the user algebraic variables just  `z`, and internally when they get collected to call them
`z_all` 

All other variables defined by the user that enter `model` do not have subscripts (except `v_global`) and
for `f_alg`, i would prefer `g_z` and all them together `g_z_all` (as it already is)

src/matlab/FiniteElement.m
```diff
@@ -282,6 +283,17 @@
...
                 end
+
+                % add user variables
+                z_user = define_casadi_symbolic(settings.casadi_symbolic_mode,...
+                                                ['z_user_' num2str(ctrl_idx-1) '_' num2str(fe_idx-1) '_' num2str(ii)],...
+                                                dims.n_z_user);
+                obj.addVariable(z_user,...
+                                'z_user',...
+                                -inf*ones(dims.n_z_user,1),...
+                                inf*ones(dims.n_z_user,1),...
```
nurkanovic, line 320: I would make it possible for the user to add lbz and ubz box constraints on the algebraic variables

examples/matlab/gene_regulation/irma_model.m
apozharski: Rename is done.

nurkanovic (approved):
Great work @apozharski 🚀

PR #35, opened 2023-02-10
Globals

apozharski:
Finally got around to adding this. Probably want an example for this as well at some point.

### DIFF
--- a/src/FiniteElement.m
+++ b/src/FiniteElement.m
@@ -563,10 +563,10 @@ classdef FiniteElement < NosnocFormulationObject
 
             for j = 1:dims.n_s
                 % Multiply by s_sot_k which is 1 if not using speed of time variable
-                [fj, qj] = model.f_x_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage);
+                [fj, qj] = model.f_x_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage, model.v_global);
                 fj = s_sot*fj;
                 qj = s_sot*qj;
-                gj = model.g_z_all_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage);
+                gj = model.g_z_all_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage, model.v_global);
                 
                 obj.addConstraint(gj);
                 if settings.irk_representation == IrkRepresentation.integral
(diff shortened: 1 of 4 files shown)


src/FiniteElement.m
```diff
@@ -563,10 +563,10 @@ function forwardSimulation(obj, ocp, Uk, s_sot, p_stage)
 
             for j = 1:dims.n_s
                 % Multiply by s_sot_k which is 1 if not using speed of time variable
-                [fj, qj] = model.f_x_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage);
+                [fj, qj] = model.f_x_fun(X_ki{j}, obj.rkStageZ(j), Uk, p_stage, model.v_global);
```
nurkanovic, line 566: why is `v_global` taken out of `model`, but not out of the local workspace (comming e.g. from unfold struct) like the other variables?

src/model_reformulation_nosnoc.m
```diff
@@ -1037,7 +1037,7 @@
     if simple_v0_guess
         v0 = zeros(n_x,1);
     else
-        [v0,~] = (f_x_fun(x0,z0_all,u0,[p_global_val; p_time_var_val(:,1)]));
+        [v0,~] = (f_x_fun(x0,z0_all,u0,[p_global_val; p_time_var_val(:,1)], v0_global));
```
nurkanovic, line 1040: hmm, the names of v0 and v0_global are a bit confusing, maybe adding a comment would be instructive.
ones are stage "velocity values", others are global time independent variables

nurkanovic (commented):
LGTM except the small comments that you can clarify to me.
Also, it is important to have an example (like the cart pole in nosnoc py) and possibly a test, to really see does it work as it should. :D

src/FiniteElement.m
apozharski: I'd like to _eventually_ tear out all (or almost all) of the `unfold_struct` usage. Inside the problem definition classes there is already none.

src/model_reformulation_nosnoc.m
apozharski: Yea a comment for now but I still don't love the `v_global` name as a whole.

nurkanovic: I am more than open to better suggestions 😁 
We were discussing for a while how to name it and we wanted to be consistent with the global user parameters.
But somewhat we run out of letters.

src/FiniteElement.m
nurkanovic: This is s good point actually.
Recently, I discussed with @FreyJo, and we agreed that actually having the prefix to where a function/variable belongs, improves the code readability.

FreyJo (approved):
LGTM!

PR #37, opened 2023-02-13
Time freezing multi

nurkanovic:
implementation of the time freezing reformulation for multiple impacts with friction.
getting rid of the previous handcrafted for only one and two contacts.


apozharski (approved):
There is some cleanup that should be done, but it isn't urgent and probably can be done in parallel with cleanup of the settings and model code.

PR #38, opened 2023-02-22
Perf improvements via Removal of Duplication of lambda/theta Calculation

nurkanovic:


apozharski:
Update is rebased and should be good to merge with something like ~20% improvement in speed on larger problems.

apozharski (approved):
LGTM

PR #39, opened 2023-02-24
added optional constraint stabilization in time-freezing

nurkanovic:
A small feature for experts that can improve the convergence.
The q dynamics of the auxiliary dynamics is per default zero and has a zero sensitivity.
However, adding a small linear dynamics term does not change the solution but adds a nonzero sensitivity and improves convergence. 
Additionally, I turned off pss_lift in default settings to zero, since it is not fully implemented for step.

### DIFF
--- a/src/time_freezing_reformulation.m
+++ b/src/time_freezing_reformulation.m
@@ -245,12 +245,12 @@ if ~time_freezing_model_exists
         f_aux_pos = []; % matrix wit all aux tan dyn
         f_aux_neg = [];
         % time freezing dynamics
-        f_q_dynamics = zeros(n_q,1);
         if settings.stabilizing_q_dynamics
-            f_aux_normal = [-settings.kappa_stabilizing_q_dynamics*J_normal.*(f_c);inv_M_aux*J_normal*a_n;zeros(1,n_contacts)];
+            f_q_dynamics = -settings.kappa_stabilizing_q_dynamics*J_normal.*(f_c);
         else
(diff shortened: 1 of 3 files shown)


src/time_freezing_reformulation.m
```diff
@@ -244,18 +245,27 @@
         f_aux_pos = []; % matrix wit all aux tan dyn
         f_aux_neg = [];
         % time freezing dynamics
-        f_aux_normal = [zeros(n_q,n_contacts);inv_M_aux*J_normal*a_n;zeros(1,n_contacts)];
+        f_q_dynamics = zeros(n_q,1);
+        if settings.stabilizing_q_dynamics
+            f_aux_normal = [-settings.kappa_stabilizing_q_dynamics*J_normal.*(f_c);inv_M_aux*J_normal*a_n;zeros(1,n_contacts)];
+        else
+            f_aux_normal = [f_q_dynamics;J_normal*a_n;zeros(1,n_contacts)];
```
apozharski, line 252: I would probably just use `zeros(n_q,1)` here and only define `f_q_dynamics` in the loop or set `f_q_dynamics` in an if statement and only assign `f_aux_normal` once.

apozharski (approved):
Looks good to me, with one minor nitpick, though I am a little shaky on the theory here.

src/time_freezing_reformulation.m
apozharski, line 252: possibly missing `inv_M_aux`

PR #42, opened 2023-06-04
Post release bugfix/performance improvements

apozharski:
s_sot extraction, options rework, and performance improvements (about 15-25% for OCPs)


src/FiniteElement.m
```diff
@@ -982,9 +983,12 @@
...
                             end
                             if (obj.fe_idx ~= 1 || ~settings.no_initial_impacts)
-                                cross_comp_pairs{2,jj+2} = [obj.w(obj.ind_Y_gap{1}), obj.w(obj.ind_lambda_normal{jj,1})]; % TODO maybe we need the other cross comps?
-                                len_pad = size(cross_comp_pairs{1,jj+2},1) - size(cross_comp_pairs{2,jj+2}, 1);
-                                cross_comp_pairs{2,jj+2} = vertcat(cross_comp_pairs{2,jj+2}, zeros(len_pad, 2));
+                                comp = [obj.w(obj.ind_Y_gap{1}), obj.w(obj.ind_lambda_normal{jj,1})]; % TODO maybe we need the other cross comps?
+                                size_comp = size(comp, 1);
+                                size_full_comp = size(cross_comp_pairs{end,jj+2});
+                                sparsity = Sparsity.rowcol([0:(size_comp-1)],[0,1],size_full_comp(1),size_full_comp(2));
+                                cross_comp_pairs{2,jj+2} = define_casadi_symbolic(settings.casadi_symbolic_mode, 'dummy', size_full_comp, sparsity);
```
FreyJo, line 990: I am a bit confused why you need a matrix here. Can you explain?

src/NosnocSolver.m
```diff
@@ -53,6 +53,7 @@
                 end
             end
 
+            settings.preprocess();
             settings.create_butcher_tableu(model); % TODO this should live somewhere else. (i.e. butcher tableu should not be in settings)
```
FreyJo, line 57: Can you move the  `create_butcher_tableau` inside `preprocess`?

examples/time_freezing_hystheresis/time_optimal_control_car/states_and_control.fig
FreyJo, line 1: I would prefer not adding the `.fig` to the repo.

examples/time_freezing_hystheresis/time_optimal_control_car/car_hystheresis_model_voronoi.m
```diff
@@ -55,7 +55,7 @@
 g_3 = norm([psi;w]-z3)^2;
 g_4 = norm([psi;w]-z4)^2;
 
-g_ind = [g_1;g_2;g_3;g_4];
+model.g_ind = [g_1;g_2;g_3;g_4];
```
FreyJo, line 58: It is not new, but I think the name can be better.
I first thought `ind` is for index, but it is indicator.

FreyJo (approved):
Looks good, added minor comments

src/FiniteElement.m
apozharski: Because Sparsity based assignment doesn't work the way I would want to (i.e I don't think there is a way to change the sparsity of an existing matrix) so we need the matrix of the correct size and populate it afterward.

examples/time_freezing_hystheresis/time_optimal_control_car/car_hystheresis_model_voronoi.m
nurkanovic: I suggest `g_Stewart` or `g_switch`, as these are the switching indicator functions from Stewart's reformulation of Filippov systems.

PR #45, opened 2023-08-11
Sigma Scaling, wall time timeouts, and fixes/changes to convergence checks.

apozharski:


src/NosnocNLP.m
```diff
@@ -110,7 +110,8 @@
             obj.sigma_p = sigma_p;
             obj.p = [sigma_p;mpcc.p];
 
-            obj.cost = mpcc.cost;            
+            obj.cost = mpcc.cost;
+            obj.objective = mpcc.objective;
```
nurkanovic, line 114: I am a bit confused about the difference between cost and objective.
I suppose one is the pure NLP cost function, and the other has the algorithmic parts like step equilibration, comp. residual and so on. Add a comment to explain the difference or provide an explicit name such as
`objective` and`objective_with_algoritmic_terms`

src/NosnocSolver.m
```diff
@@ -137,11 +137,10 @@ function set(obj, type, val)
             if strcmp(obj.solver_options.solver, 'ipopt')
                 last_stats = stats.solver_stats(end);
                 converged = 0;
-                if isfield(last_stats, 'iterations')
+                if isfield(last_stats, 'iterations') && ~isempty(last_stats.iterations)
                     inf_pr = last_stats.iterations.inf_pr(end);
                     inf_du = last_stats.iterations.inf_du(end);
-                    if inf_pr < obj.solver_options.opts_casadi_nlp.ipopt.tol && inf_du < obj.solver_options.opts_casadi_nlp.ipopt.tol ...
-                            && stats.complementarity_stats(end) < 10 * obj.solver_options.comp_tol
+                    if stats.complementarity_stats(end) < 10 * obj.solver_options.comp_tol
```
nurkanovic, line 143: why the *10?

src/NosnocSolver.m
```diff
@@ -327,6 +326,11 @@ function  print_solver_stats(obj, results, stats)
             % TODO: Possible issue raise to casadi: allow unknown fields in options passed
             if strcmp(solver_options.solver, 'ipopt')
                 opts_casadi_nlp = rmfield(solver_options.opts_casadi_nlp, 'snopt');
+                if solver_options.timeout_cpu
+                    opts_casadi_nlp.ipopt.max_cpu_time = solver_options.timeout_cpu;
+                elseif solver_options.timeout_wall
+                    opts_casadi_nlp.ipopt.max_wall_time = solver_options.timeout_wall;
```
nurkanovic, line 332: should'nt   `opts_casadi_nlp.ipopt.` be   `opts_casadi_nlp.snopt.`

src/NosnocSolverOptions.m
```diff
@@ -183,44 +191,72 @@
             switch obj.psi_fun_type
               case CFunctionType.BILINEAR
```
nurkanovic, line 192: rename to `Scholtes`

nurkanovic (changes requested):
I had a few minor comments to be adressed. Otherwise, nice job.

src/NosnocSolverOptions.m
```diff
@@ -83,6 +83,13 @@
         % Output options
         store_integrator_step_results(1,1) logical = 0
         ipopt_callback = [] % This should be a function handle that takes (model,problem,settings,ipopt_solver,results)
+
+        timeout_cpu(1,1) {mustBeReal, mustBeNonnegative} = 0;
+        timeout_wall(1,1) {mustBeReal, mustBeNonnegative} = 0;
```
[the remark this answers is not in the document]
apozharski: You could use integer max here, but a timeout of 0 already makes no sense so using it as a marker for no timeout makes sense. There is also some precedent as all of WORHP, IPOPT, and SNOPT use 0 as a placeholder for no timeout.

src/NosnocSolver.m
apozharski: No this is inside the if statement for ipopt not snopt

src/NosnocSolver.m
apozharski: That is a good question. I think this was added at some point to give a buffer because of numerics but `*10` is probably too large.

src/NosnocSolver.m
```diff
@@ -431,18 +444,44 @@ function  print_solver_stats(obj, results, stats)
...
+                                opts_casadi_nlp = rmfield(solver_options.opts_casadi_nlp, 'snopt');
+                                opts_casadi_nlp.ipopt.max_cpu_time = solver_options.timeout_cpu - stats.cpu_time_total;
+                            elseif strcmp(solver_options.solver, 'snopt')
+                                opts_casadi_nlp = rmfield(solver_options.opts_casadi_nlp, 'ipopt');
+                            end
+                            solver = nlpsol(solver_options.solver_name, solver_options.solver, casadi_nlp, opts_casadi_nlp);
+                        elseif obj.solver_options.timeout_wall
+                            casadi_nlp = struct('f', obj.nlp.cost, 'x', obj.nlp.w, 'g', obj.nlp.g, 'p', obj.nlp.p);
+                            if strcmp(solver_options.solver, 'ipopt')
+                                opts_casadi_nlp = rmfield(solver_options.opts_casadi_nlp, 'snopt');
```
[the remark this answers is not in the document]
apozharski: This line is removing the snopt field from the casadi nlp options. This is because casadi will freak out if you pass an options struct that contains unexpected fields.

src/NosnocSolverOptions.m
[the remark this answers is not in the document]
nurkanovic: okay. thanks for clarification

nurkanovic (approved):
LGTM

PR #49, opened 2023-08-23
Solver plugins, better user defined algebraics for step, and simple switching cases testing

apozharski:
Add Classes to handle the output of different solvers. Added custom time freezing flags. Performance improvements,.


src/NosnocSNOPT.m
```diff
@@ -0,0 +1,106 @@
...
+    methods
+        function solver = construct_solver(obj, nlp, solver_options, time_remaining)
+            import casadi.*
+            w = nlp.w;
+            g = nlp.g;
+            p = nlp.p;
+
+            casadi_nlp = struct('f', nlp.augmented_objective, 'x', w, 'g', g, 'p', p);
+
+            % TODO: Possible issue raise to casadi: allow unknown fields in options passed
```
FreyJo, line 42: I don't think they want to change that -> rm TODO :P

src/NosnocSolver.m
```diff
@@ -133,26 +146,15 @@ function set(obj, type, val)
...
-                    if stats.complementarity_stats(end) < 10 * obj.solver_options.comp_tol
-                        converged = 1;
-                    end
-                else
-                    converged = 0;
-                end
-            else
-                % TODO..
-                converged = [];
+        function converged = comp_tol_met(obj, stats)
```
FreyJo, line 149: I think renaming this function makes it less readable (`is_converged` -> `comp_tol_met`).
More verbose function names would be good IMO.

src/NosnocProblemOptions.m
```diff
@@ -53,20 +53,15 @@
...
 
         % TODO: Initialization is really a solver concern, but there is not really a good extraction approach yet.
         % Initialization - Stewart
         lp_initialization(1,1) logical = 0
-        initial_theta(1,1) double {mustBeReal, mustBeFinite} = 1
-        initial_lambda(1,1) double {mustBeReal, mustBeFinite} = 1
-        initial_mu(1,1) double {mustBeReal, mustBeFinite} = 1
 
         % Initialization - Step
         initial_alpha(1,1) double {mustBeReal, mustBeFinite} = 1
```
FreyJo, line 64: I think, for that, we should hardcode 0.5.

src/NosnocProblemOptions.m
```diff
@@ -53,20 +53,15 @@
...
-        initial_lambda_0(1,1) double {mustBeReal, mustBeFinite} = 1
-        initial_lambda_1(1,1) double {mustBeReal, mustBeFinite} = 1
         initial_beta(1,1) double {mustBeReal, mustBeFinite} = 1
```
FreyJo, line 65: Here I am not sure how to get rid of it, maybe add a todo, we can discuss with armin.

FreyJo (approved):
Looks good to me! Added some comments on the initialization.

src/NosnocSolver.m
apozharski: This function is no longer the only thing being checked to check for convergence hence the rename to more accurately reflect exactly what it checks.

FreyJo: I see now. Maybe rename the flag to just `flag` and the function to `complementarity_tol_met`.
I thought something like compute tolerance metric :sweat_smile:

apozharski: :100: I see.... that actually is confusing, good catch :sweat_smile:

PR #50, opened 2023-08-25
Move all discretization options from model to problem_options 

FreyJo:
These are
- T_sim
- N_sim
- h_sim
- h
- h_k
- T

### DIFF
--- a/src/NosnocProblemOptions.m
+++ b/src/NosnocProblemOptions.m
@@ -37,7 +37,7 @@ classdef NosnocProblemOptions < handle
 
         h % Step size
         h_k % Finite element step size
-        T = 1.0 % Terminal time % TODO: why is there also T_val, do we need both?
+        T % Terminal time % TODO: why is there also T_val, do we need both?
 
         % descritization
         N_stages(1,1) {mustBeInteger, mustBePositive} = 1;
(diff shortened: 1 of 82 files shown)


src/NosnocProblemOptions.m
```diff
@@ -90,13 +98,13 @@
...
         impose_terminal_phyisical_time(1,1) logical = 1
-        s_sot0(1,1) double {mustBeReal, mustBePositive} = 1
-        s_sot_max(1,1) double {mustBeReal, mustBePositive} = 25
-        s_sot_min(1,1) double {mustBeReal, mustBePositive} = 1
-        S_sot_nominal(1,1) double {mustBeReal, mustBePositive} = 1
+        s_sot0(1,1) double {mustBePositive} = 1
+        s_sot_max(1,1) double {mustBePositive} = 25
+        s_sot_min(1,1) double {mustBePositive} = 1
+        S_sot_nominal(1,1) double {mustBePositive} = 1
         rho_sot(1,1) double {mustBeReal, mustBeNonnegative} = 0
```
apozharski, line 105: We can drop the `mustBeReal` from here as well, but not super important.

apozharski (approved):
One comment, but I can handle it later when we do a cleanup pass.
LGTM :rocket:

Issue #55, opened 2023-09-20
The main_throwing_ball.m example is not working, the setting is wrong?


chenglong0791:
clear all
clc
close all
import casadi.*
%% model parameters
e = 0.2; u_max = 9; beta = 0.0; mu=0;
%% NOSNOC settings
problem_options = NosnocProblemOptions();
solver_options = NosnocSolverOptions();  %% Optionally call this function to have an overview of all options.

problem_options.n_s = 3; 
solver_options.mpcc_mode = 'Scholtes_ineq';
solver_options.homotopy_update_rule = 'superlinear';
% problem_options.step_equilibration = 'direct_homotopy';
problem_options.use_fesd = 1;
problem_options.time_freezing = 1;

problem_options.local_speed_of_time_variable = 1;
problem_options.stagewise_clock_constraint = 0;
problem_options.nonsmooth_switching_fun = 0;
problem_options.pss_lift_step_functions = 0;

problem_options.no_initial_impacts = 0;
% solver
solver_options.opts_casadi_nlp.ipopt.max_iter = 1e5;
solver_options.opts_casadi_nlp.ipopt.tol = 1e-7;
solver_options.opts_casadi_nlp.ipopt.acceptable_tol = 1e-5;
solver_options.use_previous_solution_as_initial_guess = 0;
solver_options.N_homotopy = 6;
solver_options.print_level = 3;
solver_options.comp_tol = 1e-6;
solver_options.elastic_scholtes = 1;
%% model parameters
g=9.81;
q_target = [4;0.5];

%% solver
model = NosnocModel();
model.mu_f = mu;

problem_options.T = 4; 
problem_options.N_stages = 20; 
problem_options.N_finite_elements = 3;
% model equations
q = SX.sym('q',2);
v = SX.sym('v',2); 
u = SX.sym('u',2);

model.lbx = [0;-0.001;0;-inf]; model.ubx = [4;inf;inf;inf];
model.x0 = [0;0.5;0;0];
model.x = [q;v]; model.u = u; model.e = e ;
model.f_c = q(2)-0;
model.f_v = [u-[0;g]-beta*v*sqrt(v(1)^2^2+v(2)^2+1e-3)]; 
model.J_tangent = [1; 0];
model.dims.n_dim_contact = 2;

% Objective and constraints
model.f_q = u'*u; model.f_q_T = 100*v'*v;
model.g_path = u'*u-u_max^2;
model.g_terminal = q-[q_target];

[...]

PR #56, opened 2023-09-20
Addressing Issue with Throwing ball example.

apozharski:
Fix throwing ball example and plotting. Addresses #55.


nurkanovic (approved):
lgtm

PR #57, opened 2023-09-20
Fix nu function and step equilibration options. Implement CLS nu function

apozharski:
Step equilibration (that relies on $\nu$) was broken during implementation of CLS/splitting of NLP and MPCC generation. This attempts to bring that back into working order.

### DIFF
--- a/src/FiniteElement.m
+++ b/src/FiniteElement.m
@@ -1070,8 +1070,8 @@ classdef FiniteElement < NosnocFormulationObject
                         pi_discont = sigma_discont_B .* sigma_discont_F;
                         nu = pi_cont + pi_discont;
                     case DcsMode.CLS
-                        f_c_F = cellfun(@(x) obj.model.f_c_fun(obj.w(x)), obj.ind_x, 'uni', false);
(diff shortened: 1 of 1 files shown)


src/FiniteElement.m
```diff
@@ -1037,19 +1037,115 @@
...
+                                        sigma_beta_F = sum2(horzcat(beta_F{:}));
+                                        sigma_beta_B = sum2(horzcat(beta_B{:}));
+
+                                        pi_xi_p = sigma_xi_p_F .* sigma_xi_p_B;
+                                        pi_xi_n = sigma_xi_n_F .* sigma_xi_n_B;
+                                        pi_beta = sigma_beta_F .* sigma_beta_B;
+
+                                        zeta(ii) = (sigma_f_c_B + sigma_f_c_F) + (sum(pi_xi_p + pi_xi_n) + pi_beta);
+                                    else
+                                        % TODO: Verify
```
apozharski, line 1122: I may be missing things in the polyhedral implementation. Need to think whether we need the left boundary points here for switches that switch while slipping (only $\gamma_{n,0} = 0$?).

FreyJo (approved):
Quite important to have the Nurkanovic function fixed! :+1: 
I guess there is no simple unified way to do that.
But the way it is done now allows to easily compare with the publications.
Obviously the TODOs especially for the subsystems would be good to address, but are not trivial.

PR #62, opened 2024-03-27
 Initial extraction of NosnocSolver to work for generic MPCCs

apozharski:
First step to a better Nosnoc. This PR contains the following changes:
- Introduction of an `nlpsol` style interface for MPCCs in the form of `mpccsol`. 
- Extraction of a generic form of `NosnocSolver` into `nosnoc.solver.RelaxationSolver` to provide a generic interface to the homotopy solver.
- A simple example of a generic MPCC solved with `mpccsol`
- `NosnocSolver` now uses `mpccsol` as its backend
-  Examples and tests have been updated to use `nosnoc.solver.Options`
- `NosnocNLP` removed and nlpsolver plugin classes moved to `nosnoc.solver`
- ~~Test action updated to automatically pull latest version of `vdx`~~ `vdx` as a sub-module.

This implementation is yet missing some of the stationarity check and polishing steps but this is planned to be added.

This PR also contains the first use of `vdx` (https://github.com/apozharski/vdx) for index tracking which will hopefully be extensively used in the cleanup of the Nosnoc core.


+nosnoc/+solver/+plugins/Ipopt.m
```diff
@@ -1,6 +1,6 @@
 % BSD 2-Clause License
 
-% Copyright (c) 2022, Armin Nurkanović, Jonathan Frey, Anton Pozharskiy, Moritz Diehl
+% Copyright (c) 2024, Armin Nurkanović, Jonathan Frey, Anton Pozharskiy, Moritz Diehl
```
FreyJo, line 3: I would always use the same header, including the year.

.github/workflows/test_build.yml
```diff
@@ -10,7 +10,18 @@ jobs:
...
+      with:
+        path: nosnoc
+
+    # Checkout dependency
+    # TODO(@anton) make a release of vdx once it is stable
+    - name: Checkout vdx
+      uses: actions/checkout@v3
+      with:
+        repository: apozharski/vdx
+        path: vdx
```
FreyJo, line 24: What about making `vdx` a submodule?

FreyJo (approved):
The changes look good to me!

nurkanovic:
There are some ambiguities/overloading that need to be resolved in the MPCC method - as they will for sure confuse anyone outside (but probably also many inside) syscop. 

Some things are confused and used interchangeably, i.e.  the enums from 
CFunctionType, MpccMode and MpccRelaxationMethod

Also grouping solution methods into classes like now into Relaxation being the only one creates an additional layer of complexity and makes also code often longer. I would be in favor to not using it. Scholtes will be anyway the default, an mpcc-educated user knows that it is a homotopy relaxation method.

Now  we have Scholtes_ineq or Scholtes_eq in the MPCC methods and in C_functions we have Scholtes. 
And then one can pick in the options/setting both the mpcc method, but then somewhat overwrite with C_functions.  Even I am confused by this, even though I know the math behind it in great detail.

I suggest the following: 
 - we distinguish between MPCC methods and they have the names as in the literature, e.g. Scholtes_ineq method, Kazow-Schwartz, and so on (this would make CFunfctionType not needed anymore).
 We would add to this list also other new methods that are currently in development (even if they are not of the relaxation type).  
 - Next we have how we treat the comp constraints (currently MPCCRelaxationMode), the default should be ineq as the others would usually work less good - in fact we remove this to be chosen manually but only via the list of methods (i would not even support eq and two_sided for most of the c-funs, as it is useless) 
 - Then we have to look at HomotopyParameterSteering (relaxation, ell1, ell_inf). 
 In line with Table 1 in https://arxiv.org/pdf/2312.11022.pdf#page=12.43 

To make the interface even simpler I suggest to make just list MPCCMethods (currently called MPCCMode)
which is composed of the cartesian product of  Cfunctiontype-HomotopyParameterSteering-MPCCRelaxationMode.  We do not need to list all, because we know that some combinations (especially with eq are just bad, and keep maybe the ones for which we know that they are actually good)

New methods would be simply added to this list. They might need an homotopy but be a separate function that is called. the others can share the homotopy loop (when one actually starts reorganizing this, it will become clear what is the best way to do it)

examples/solve_generic_mpcc/generic_mpcc.m
```diff
@@ -0,0 +1,30 @@
+close all;
+clear all;
+import casadi.*;
+import nosnoc.solver.mpccsol;
+
+mpccsol_opts = struct;
```
nurkanovic, line 6: this line not needed

examples/solve_generic_mpcc/generic_mpcc.m
```diff
@@ -0,0 +1,30 @@
...
+opts = nosnoc.solver.Options();
+mpccsol_opts.relaxation = opts;
```
nurkanovic, line 8: as discussed in the big comment, not happy with the additional layer of relaxation, as currently, we have only this.
and the other methods that come will have most of the settings the same. 
we can fall back to another layer if it becomes necessary.

examples/solve_generic_mpcc/generic_mpcc.m
```diff
@@ -0,0 +1,30 @@
...
+mpcc.G.x(0) = {x(1)};
+mpcc.H.x(0) = {x(2)};
+
+mpcc_struct = struct;
+mpcc_struct.x = mpcc.w.sym;
+mpcc_struct.g = mpcc.g.sym;
+mpcc_struct.p = mpcc.p.sym;
+mpcc_struct.G = mpcc.G.sym;
+mpcc_struct.H = mpcc.H.sym;
+mpcc_struct.f = mpcc.f;
```
nurkanovic, line 23: as discussed in the vdx comments,  the vdx Problem could possibly contain this struct and would shorten the manual extraction of the symbolics.

nurkanovic:
Some further todos/remarks: 
- I have added some simple mpec examples in the folder solver_generic_mpcc. 
There one can see that it is not allowed to not provide ``mpcc_struct.p`` or ``mpcc_struct.g``, if the problem is not parametric or  does not have further constraints, resp.
- There are many options \nosnoc\+nosnoc\+solver\Options that are not related to a generic mpcc, e.g., elastic_scholtes(1,1) logical (what is this doing?), everying below the comment "        % Integrator Specific" , multiple_solvers,         % Output options
- The standalone mpcc solver has no verbose
- The psi_fun settings should be simplfied according to my previous long comment, where method (Scholtes), and mode (eq,ineq,two-sided) are unified, and homotopy steering is left (elastic_ell1,...)
- vdx is not needed to define an mpcc, as this is the only case where we use it in the frond end, i suggest to move this example to the vdx respository, and keep the simple ones that i have added
- Playing a bit with the options in the examples might be helpful for users as a template. 
- calculate_stationarity_type = true, does nothing on my machine. maybe it would be useful to do this via Highs instead of gurobi, which is shipped via casadi anyway (but i would let this be a separate issue/pr)

+nosnoc/+solver/RelaxationSolver.m
```diff
@@ -146,11 +154,7 @@
 
         function violation = compute_constraint_violation(obj, w, g)
             nlp = obj.nlp;
-            ubw_violation = max(max(w - nlp.w.mpcc_w(0).ub), 0);
-            lbw_violation = max(max(nlp.w.mpcc_w(0).lb - w), 0);
-            ubg_violation = max(max(g - nlp.g.mpcc_g(0).ub), 0);
-            lbg_violation = max(max(nlp.g.mpcc_g(0).lb - g), 0);
-            violation = max([lbg_violation, ubg_violation, lbw_violation, ubw_violation]);
+            violation = max([nlp.g.mpcc_g().violation; nlp.w.mpcc_w().violation]);
```
nurkanovic, line 188: would you mind adding one line of comment explaing this vdx synthax here (what is the output of violation

+nosnoc/+solver/RelaxationSolver.m
```diff
@@ -54,34 +60,36 @@
...
+                
                 nlp = vdx.Problem('casadi_type', casadi_symbolic_mode);
-                nlp.w.mpcc_w(0) = {mpcc.x};
-                nlp.p.mpcc_p(0) = {mpcc.p};
-                nlp.g.mpcc_g(0) = {mpcc.g};
-                nlp.p.sigma_p(0) = {{'sigma_p', 1}, 0, inf, opts.sigma_0};
+                nlp.w.mpcc_w = {mpcc.x};
+                nlp.p.mpcc_p = {mpcc.p};
+                nlp.g.mpcc_g = {mpcc.g};
+                nlp.p.sigma_p = {{'sigma_p', 1}, opts.sigma_0};
```
nurkanovic, line 80: can you please comment that an nlp is created out of the mpec symbolic functions.
nlp  - is a vdx object, nlp.w - is a variablevectors (or primal vectors, whatever is the official name), 
mpcc_w  is a variable group (right name?) that contains all variables of the initial mpcc. New vars might be added in the nlp due to the reformulation. 
(these comments will help one to get into reading vdx code faster, after some sytntax was forgotten)

+nosnoc/+solver/RelaxationSolver.m
```diff
@@ -54,34 +60,36 @@
...
 
                 % Create relaxation slacks/parameters
                 switch opts.elasticity_mode
                   case ElasticityMode.NONE
-                    sigma = nlp.p.sigma_p(0);
+                    sigma = nlp.p.sigma_p();
                   case ElasticityMode.ELL_INF
-                    nlp.w.s_elastic(0) = {{'s_elastic', 1}, opts.s_elastic_min, opts.s_elastic_max, opts.s_elastic_0};
-                    sigma = nlp.w.s_elastic(0);
+                    nlp.w.s_elastic = {{'s_elastic', 1}, opts.s_elastic_min, opts.s_elastic_max, opts.s_elastic_0};
```
nurkanovic, line 89: add comment in code: adding elastic mode variable to the existing mpcc variables

+nosnoc/+solver/RelaxationSolver.m
```diff
@@ -54,34 +60,36 @@
...
+                nlp.g.mpcc_g = {mpcc.g};
+                nlp.p.sigma_p = {{'sigma_p', 1}, opts.sigma_0};
                 nlp.f = mpcc.f;
                 n_c = size(mpcc.G);
 
                 % Create relaxation slacks/parameters
                 switch opts.elasticity_mode
                   case ElasticityMode.NONE
-                    sigma = nlp.p.sigma_p(0);
+                    sigma = nlp.p.sigma_p();
```
nurkanovic, line 87: nlp.p.sigma_p( )  means that sigma is a parameter/variable that is not time depended and has no indicies - please add the comment

+nosnoc/+solver/RelaxationSolver.m
apozharski: Comment added.

+nosnoc/+solver/RelaxationSolver.m
apozharski: Comment added.

+nosnoc/+solver/RelaxationSolver.m
apozharski: Comment added on violation.

+nosnoc/+solver/RelaxationSolver.m
apozharski: Added Comment

nurkanovic:
I have edited the mpec_ex1 and ell_1 seems to be broken now. I suggest to
- to make a test out of this example to test the most basic mpccsol options, like few solver, elasticity mode
- to change the name of elasticity_mode to homotopy_steering_strategy to be in line with Table 1 of https://arxiv.org/pdf/2312.11022.pdf and to rename NONE to DIRECT 
- update the installation part in the README as due to vdx a recursive update and installation script must be run

Btw, is also the new parameter update (which i like a lot btw) tested for the time-varying case? (solving a single ocp instance with this kind of parameters would be enough)

I hope these are my last comments on this PR. :D

Issue #79, opened 2024-04-15
Add verbose output to `mpccsol`


apozharski:
Based on #62

PR #88, opened 2024-04-18
Fix verbose output for `mpccsol`

apozharski:
Address #79.


nurkanovic (approved):
Looks fine 👍 - it only remains, as discussed, to get the alignment for different terminals consistent

Issue #80, opened 2024-04-15
Re-implement calculation of stationarity type via TNLP/RNLP and LCP->MILP


apozharski:
Based on #62 

Re-implement functionality to calculate the type of stationarity point in `RelaxationSolver`.

Issue #78, opened 2024-04-15
Finish re-implementing polishing step.


apozharski:
Based on conversation in #62.

Re-implement the fixed active set NLP extraction for finding a polished solution.

PR #89, opened 2024-04-18
Stationarity calculation and polishing

apozharski:
Addressing #78 #80


+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +426,443 @@ function print_infeasibility(obj, results)
...
+        function [polished_w, res_out, stat_type, n_biactive] = calculate_stationarity(obj, results, exitfast, lifted)
+            import casadi.*
+            stat_type = "?";
+            nlp = obj.nlp;
+            mpcc = obj.mpcc;
+
+            % This method generates a TNLP (tightened NLP) using the results of a homotopy solver.
+            % In order to do so we first identify the active set of the complementarity constraints which involves identifying which
+            % branch of the complementarity each element of G(w) and H(w) are, or whether the solution is biactive at that point.
+            % For a more in depth explanation see the PhD theses of Armin Nurkanovic or Alexandra Schwartz
```
nurkanovic, line 471: We can be more specific with Pointing to Section 2.3 of https://publications.syscop.de/Nurkanovic2023f.pdf
and to https://arxiv.org/abs/2312.11022

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +426,443 @@ function print_infeasibility(obj, results)
...
+            
+            w_orig = results.x(nlp.ind_map);
+            lam_x = full(results.lam_x(nlp.ind_map));
+            lam_g = full(results.lam_g);
+            lam_g(nlp.ind_g_comp) = []; % TODO: also elastic scholtes constraint
```
nurkanovic: what is this? I don' understand what is meant by elastic scholtes constraints and how it is relevant?

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +426,443 @@ function print_infeasibility(obj, results)
...
+
+            G = mpcc.G_fun(mpcc.w, mpcc.p);
+            H = mpcc.H_fun(mpcc.w, mpcc.p);
+            G_old = full(mpcc.G_fun(w_orig, obj.p_val(2:end)));
+            H_old = full(mpcc.H_fun(w_orig, obj.p_val(2:end)));
+            if obj.solver_options.elasticity_mode == ElasticityMode.NONE
```
nurkanovic: didnt we remove the elasticit mode option?

+nosnoc/+solver/MpccSolver.m
```diff
@@ -421,7 +422,9 @@
                     end
                 end
                 if stat_type ~= "?"
-                    mpcc_results.W = [mpcc_results.W,polished_w];
+                    mpcc_results.x = polished_w;
```
nurkanovic, line 425: why not call it w_polished, as we always have w_{whatever_is_discribing_it}

+nosnoc/+solver/MpccSolver.m
apozharski: I believe you are looking at an old commit here. This no longer exists.

+nosnoc/+solver/MpccSolver.m
```diff
@@ -402,6 +403,34 @@
             stats.success = stats.converged;
             stats.constraint_violation = obj.compute_constraint_violation(mpcc_results.x, mpcc_results.g);
 
+            if opts.calculate_stationarity_type
+                if last_iter_failed || ~obj.complementarity_tol_met(stats) || timeout
+                    stat_type = "?";
+                    disp("Not checking stationarity due to failure of homotopy to converge");
```
nurkanovic, line 410: homotopy to converge -> homotopy to converge.

+nosnoc/+solver/MpccSolver.m
```diff
@@ -402,6 +403,37 @@
...
+                    if ~b_stat
+                        [polished_w, res_out, stat_type, n_biactive] = obj.calculate_stationarity(true, true);
+                    else
+                        stat_type = "B";
+                    end
+                end
+                if stat_type ~= "?"
+                    mpcc_results.x = polished_w;
+                    mpcc_results.f = full(f_mpcc_fun(polished_w, nlp.p.mpcc_p().val));
+                    % TODO (@anton) also recalculate multipliers and g?
```
nurkanovic, line 427: the TNLP and the homotopy NLP are two different NLPs, the multipliers are completley different (except maybe in the case when the point i S stationary and the bilinear term is inactive)

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +458,415 @@ function print_infeasibility(obj, results)
...
+            w_orig = nlp.w.mpcc_w().res;
+            lam_x = nlp.w.mpcc_w().mult;
+            lam_g = nlp.g.mpcc_g().mult;
+
+            G = mpcc.G;
+            H = mpcc.H;
+            G_old = full(obj.G_fun(w_orig, nlp.p.mpcc_p().val));
+            H_old = full(obj.H_fun(w_orig, nlp.p.mpcc_p().val));
+            if obj.opts.homotopy_steering_strategy == HomotopySteeringStrategy.DIRECT
+                a_tol = 1*sqrt(obj.opts.comp_tol);
```
nurkanovic, line 482: add comment why sqrt and why 1*

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +458,415 @@ function print_infeasibility(obj, results)
...
+                polished_w = [];
+                res_out = []
+                stat_type = "S";
+                disp("Converged to S-stationary point")
+                return
+            end
+            mindists = max(G_old, H_old);
+            [~, min_idx] = sort(mindists);
+
+            if lifted
```
nurkanovic, line 499: rename lifted to slightly more descriptive name, e.g., complementarity_constraints_lifted

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +458,415 @@ function print_infeasibility(obj, results)
...
+                        'lbg', lbg,...
+                        'ubg', ubg,...
+                        'lam_g0', lam_g_aug,...
+                        'lam_x0', lam_x_aug,...
+                        'p', nlp.p.mpcc_p().val);
+                    res_out = tnlp_results;
+                    if tnlp_solver.stats.success
+                        converged = true;
+                    else
+                        n_biactive = n_biactive + 1;
```
nurkanovic, line 593: ?

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +458,415 @@ function print_infeasibility(obj, results)
...
+                    end 
+                    %converged = true;
+                end
+                w_star = full(tnlp_results.x);
+                w_star_orig = w_star(ind_mpcc);
+                max(abs(w_init - w_star))
+                [~,sort_idx] = sort(full(abs(w_init - w_star))); % sorted by difference from original stationary point
```
nurkanovic, line 600: for what is this idx used for? why sort the diffrence?

+nosnoc/+solver/MpccSolver.m
```diff
@@ -426,5 +458,415 @@ function print_infeasibility(obj, results)
...
+            solver_settings.BigM = 1e2;
+
+            if ~n_biactive
+                solver_settings.fixed_y_lpcc = y_lpcc;
+            end
+            %% The problem
+            rnlp = struct('x', x, 'f', f, 'g', g, 'comp1', lift_G, 'comp2', lift_H, 'p', p);
+            solver_initalization = struct('x0', x0, 'lbx', lbx, 'ubx', ubx,'lbg', lbg, 'ubg', ubg, 'p', p0, 'y_lpcc', y_lpcc);
+            solution = b_stationarity_oracle(rnlp,solver_initalization,solver_settings);
+            cpu_time_filterSMPCC2 = toc;
```
nurkanovic, line 865: change name of the variable to more descriptive one

+nosnoc/+solver/MpccSolver.m
apozharski: Sure but in this case I mean the multipliers on G/H which may be able to be correctly calculated in the case of S-stationarity. This would allow MpccSolver to return correct multipliers for the MPCC Lagrangian.

+nosnoc/+solver/MpccSolver.m
apozharski: Because we generally have `a*b-sigma` (at least for Scholtes). In this case the intersection of the level set at sigma and a=b is `sqrt(sigma)`. Added a comment.

+nosnoc/+solver/MpccSolver.m
apozharski: This is part of the algorithm that adds more (possibly ambiguous) points to the biactive set. In some cases the initial guess at a biactive set may be inaccurate and some ambiguous points are also biactive (this for example often happens in time freezing problems). As such we sequentially add points to the biactive set and try to solve the TNLP again.

+nosnoc/+solver/MpccSolver.m
apozharski: This was debugging info. Removing.

+nosnoc/+solver/MpccSolver.m
apozharski: Removed as this is vestigial.

+nosnoc/+solver/MpccSolver.m
nurkanovic: I see, a description in the comment would be fine. 
And when this procedure stops should be explained.

+nosnoc/+solver/MpccSolver.m
nurkanovic: Strictly speaking, if a TNLP has a solution then the multipliers are MPCC Lag. multipliers. If MPEC-SMFCQ holds they are even unique, but they dont have to be (especially for the weaker CQs) and an NLP solver will return a multiplier, depending on the internal reglarization.
If it S stat. ,then one can compute "ture" multipliers of the MPCC.

nurkanovic:
There are few things that would be very useful. 
Currently the stationarity checks only one type, however, after the TNLP is found and multipliers are computed it would be useful to have:
- on the one hand, to determine the multiplier-based stationarity type (W,A,C,M,S), and
- on the other hand to additionally verify is is on top of that a B-stationarity point and to have both information (if it is S, one can omit checking B, as it is implied) 
- it would be nice to have methods on the mpcc/solver .check_multiplier_based_stationarity() and .check_if_b_stationary(), to check a posteriori the type of stationarity (even if it is turned off in the settings) - essentially the same code would be called.
- comp_tol should be renamed to complementarity_tol to be more instructive.

Further remarks:
- I have added a small example, with comments in it. This example breaks several options, and the stationarity check is not working correctly, i.e., delivers wrong and misleading results. 
- by an option, (default off), we may hide the verbose of the B-stat check from the user?
- Sometimes in the disp/fprintf report a "." is missing at the end of the sentence. 
- There is some plotting of multiplier-based stationary points but it does not seem to do anything,

+nosnoc/+solver/MpccSolver.m
```diff
@@ -41,7 +41,7 @@
         ind_scalar_H
         ind_nonscalar_G
         ind_nonscalar_H
-        ind_map_G
+        ind_map_G % what is map_G?
```
apozharski, line 44: This is `(ind_map)_G` not `ind_(map_G)`. mildly confusing naming maybe but essentially it is used to track the mapping in case of lifting.

+nosnoc/+solver/MpccSolver.m
```diff
@@ -198,14 +200,16 @@
...
-            error('Concatenation not supported')
+            error('Concatenation not supported.')
         end
 
         function varargout = size(obj,varargin)
             varargout = 1;
         end
         
         function ind = end(obj,k,n)
+            % TODO end is a matlab keyword, maybe use a different name?
```
apozharski, line 211: This is a required overload for overloading paren indexing. In cases where concatenation is not supported returning 1 is correct.

+nosnoc/+solver/MpccSolver.m
nurkanovic: this was my guess, please add a comment in the code. a reminder is helpful

+nosnoc/+solver/MpccSolver.m
nurkanovic: please add this comment to the code

examples/generic_mpcc/mpec_no_s_stationarity.m
```diff
@@ -8,17 +8,20 @@
...
-% mpccsol_opts.homotopy_steering_strategy = "Direct";
+mpccsol_opts.homotopy_steering_strategy = "Direct";
 % mpccsol_opts.homotopy_steering_strategy = "ELL_1";
-mpccsol_opts.homotopy_steering_strategy = "ELL_INF"; 
+% mpccsol_opts.homotopy_steering_strategy = "ELL_INF"; 
 mpccsol_opts.calculate_stationarity_type = 1;
 
-mpccsol_opts.complementarity_tol = 1e-12;
+
+mpccsol_opts.complementarity_tol = 1e-20;
```
apozharski, line 17: I am not surprised it doesn't work. Ideally we should throw an error if someone selects a complementarity tolerance this small as it is lower than machine precision (about `1e-15`). Though I will check whether there is a structural issue here.

apozharski: Yea the homotopy fails to converge as the best complementarity residual it can get to is `1.58e-17` and never reaches `1e-20 ` which we view as a failure of `mpccsol` to converge. I think this is a willnotfix.

nurkanovic (approved):
everything is fine now. gj 🚀

PR #90, opened 2024-05-09
Full nosnoc pipeline for pss using the Stewart reformulation

apozharski:


+nosnoc/+dcs/stewart.m
```diff
@@ -0,0 +1,118 @@
...
+        z_all
+
+        f_x
+        g_Stewart
+
+        sys_idx
+
+        dims
+
+        % TODO some of these should be in base perhaps
```
nurkanovic, line 21: All 
f_x_fun, 
f_q_fun, 
 g_z_fun, g_alg_fun, It is not clear which of the equations is for the user algebraic and which is DCS algebraic (like e' theta = 1)
g_path_fun
g_comp_path_fun
g_terminal_fun
f_q_T_fun
f_lsq_x_fun
f_lsq_u_fun
f_lsq_T_fun

are generic and could be in the base clase. Why not put them there?

Stewart specific are 
        g_switching_fun
        g_Stewart_fun
        lambda00_fun

+nosnoc/+dcs/stewart.m
```diff
@@ -0,0 +1,118 @@
...
+
+        function generate_equations(obj, opts)
+            import casadi.*
+            model = obj.model;
+            dims = obj.dims;
+            
+            obj.f_x = zeros(dims.n_x,1);
+            for ii = 1:dims.n_sys
+                obj.f_x = obj.f_x + model.F{ii}*obj.theta_sys{ii};
+                obj.g_Stewart{ii} = -model.S{ii}*model.c{ii};
```
nurkanovic, line 80: what if g_Stewart given, and model.S and model.c are not, e.g. in  Voronoi type PSS?

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+        end
+
+        function create_variables(obj)
+            dims = obj.dcs.dims;
+            dcs = obj.dcs;
+            model = obj.model;
+            opts = obj.opts;
+
+            obj.p.rho_h_p = {{'rho_h_p',1}, 1};
+            obj.p.rho_terminal_p = {{'rho_terminal_p',1}, 1};
```
nurkanovic, line 25: just a remark: when we add generic constraint relaxation, then the penality parameter should be defined still here and then fed to vdx, or implicitly defined by vdx, when one defines a constraints a specifics that it must be relaxed.
also in this case, the parameter might be (1/sigma).

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+            % 2d vars
+            if obj.opts.use_fesd
+                ubh = (1 + opts.gamma_h) * h0;
+                lbh = (1 - opts.gamma_h) * h0;
+                if opts.time_rescaling && ~opts.use_speed_of_time_variables
+                    % if only time_rescaling is true, speed of time and step size all lumped together, e.g., \hat{h}_{k,i} = s_n * h_{k,i}, hence the bounds need to be extended.
+                    ubh = (1+opts.gamma_h)*h0*opts.s_sot_max;
+                    lbh = (1-opts.gamma_h)*h0/opts.s_sot_min;
+                end
+                obj.w.h(1:opts.N_stages,1:opts.N_finite_elements(1)) = {{'h', 1}, lbh, ubh, h0};
```
nurkanovic, line 58: opts.N_finite_elements(1) is a bit dirty. 
I know currently the different finite elements per stage (*) is broken - but maybe we make opts.N_finite_elements just a scalar again, and allow it to be a vector later when (*) is supported again - or we start supporting (*) right now

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+            obj.w.mu(0,0,opts.n_s) = {{'mu', dims.n_mu},0,inf};
+            obj.w.mu(1:opts.N_stages,1:opts.N_finite_elements(1),1:(opts.n_s+rbp)) = {{'mu', dims.n_mu},0,inf};
+
+            % Handle x_box settings
+            if ~opts.x_box_at_stg && opts.irk_representation ~= IrkRepresentation.differential
+                obj.w.x(1:opts.N_stages,1:opts.N_finite_elements(1),1:(opts.n_s+rbp-1)).lb = -inf*ones(dims.n_x, 1);
+                obj.w.x(1:opts.N_stages,1:opts.N_finite_elements(1),1:(opts.n_s+rbp-1)).ub = inf*ones(dims.n_x, 1);
+            end
+
+            if ~opts.x_box_at_fe
```
nurkanovic, line 100: Maybe we should have enums how fine to discretize path and box constraints, instead of separate options with true false for every possiblity

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                obj.w.x(1:opts.N_stages,1:(opts.N_finite_elements(1)-1),opts.n_s+rbp).lb = -inf*ones(dims.n_x, 1);
+                obj.w.x(1:opts.N_stages,1:(opts.N_finite_elements(1)-1),opts.n_s+rbp).ub = inf*ones(dims.n_x, 1);
+            end
+        end
+
+        function forward_sim_constraints(obj)
```
nurkanovic, line 106: rename forward_sim -> direct_transcription

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                if obj.opts.use_speed_of_time_variables && opts.local_speed_of_time_variable
+                    s_sot = obj.w.sot(ii);
+                elseif obj.opts.use_speed_of_time_variables
+                    s_sot = obj.w.sot();
+                else
+                    s_sot = 1;
+                end
+
+                sum_h = 0;
+                for jj=1:opts.N_finite_elements(ii)
```
nurkanovic, line 148: here N_finite_elemnts  as vector seems to be supported - maybe we addapt it above as well

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                    if obj.opts.use_fesd
+                        h = obj.w.h(ii,jj);
+                        sum_h = sum_h + h;
+                    elseif opts.time_optimal_problem && ~opts.use_spee_of_time_variables
+                        h = 0; % TODO
+                    else
+                        h = h0;
+                    end
+                    switch opts.irk_representation
```
nurkanovic, line 157: a general comment on naming. 
we have now the name IRK everywhere (e.g. A_irk, B_irk).. however we also have Explicit_RK as an option (honestly not used often, but still a valid choice). Hence, it not aesthetic to have IrkMode = Explicit_RK, 
I suggest to rename all IRK to RK as this covers both IRK and ERK methods.

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                      case IrkRepresentation.integral
```
nurkanovic, line 158: add a brief comment in code, integral  = stage variables are states

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                            mu_ijk = obj.w.mu(ii,jj,opts.n_s+1);
+
+                            obj.g.dynamics(ii,jj,opts.n_s+1) = {x_ijk - x_ij_end};
+                            obj.g.switching(ii,jj,opts.n_s+1) = {dcs.g_switching_fun(x_ijk, z_ijk, lambda_ijk, mu_ijk, v_global, p)};
+                            obj.g.z(ii,jj,opts.n_s+1) = {dcs.g_z_fun(x_ijk, z_ijk, ui, v_global, p)};
+                        end
+                        if ~opts.g_path_at_stg && opts.g_path_at_fe
+                            obj.g.path(ii,jj) = {dcs.g_path_fun(x_ijk, z_ijk, ui, v_global, p), model.lbg_path, model.ubg_path};
+                        end
+                      case IrkRepresentation.differential
```
nurkanovic, line 206: add a brief comment in code, integral  = stage variables are state derivatives

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+
+            % Terminal constraint
+            if opts.relax_terminal_constraint_homotopy
+                error("Currently unsupported")
+            end
+            g_terminal = dcs.g_terminal_fun(obj.w.x(opts.N_stages,opts.N_finite_elements(opts.N_stages),opts.n_s+rbp),...
+                obj.w.z(opts.N_stages,opts.N_finite_elements(opts.N_stages),opts.n_s+rbp),...
+                v_global, p_global);
+            switch opts.relax_terminal_constraint
+              case 0 % hard constraint
```
nurkanovic, line 375: we should use an ENUM insetaed of 0,1,2,3 (the same key words can be then used when we have latter the general constraint relaxations), set e.g. via different options.

opts.terminal_constraint_relaxation_type = 'None',  'Ell_1', 'Ell_2', 'Ell_inf',  'None' is default

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                obj.w.s_terminal_ell_inf = {{'s_terminal_ell_inf', 1}, 0, inf, 1e3};
+
+                g_terminal = [g_terminal-model.lbg_terminal-obj.w.s_terminal_ell_inf();
+                    -(g_terminal-model.ubg_terminal)-obj.w.s_terminal_ell_inf()];
+                obj.g.terminal = {g_terminal, -inf, 0}
+                obj.f = obj.f + obj.p.rho_terminal_p()*obj.w.s_terminal_ell_inf();
+            end
+        end
+
+        function generate_complementarities(obj)
```
nurkanovic, line 401: rename, e.g. generate_complementarty_constraints

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+            import casadi.*
+            opts = obj.opts;
+            dcs = obj.dcs;
+            model = obj.model;
+            % Do Cross-Complementarity
+
+            rbp = ~opts.right_boundary_point_explicit;
```
nurkanovic, line 408: why a negation of a setting, might confuse while reading the code later

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+
+                            obj.G.standard_comp(ii,jj, kk) = {lambda_ijk};
+                            obj.H.standard_comp(ii,jj, kk) = {theta_ijk};
+                        end
+                    end
+                end
+            end
+        end
+
+        function step_equilibration(obj)
```
nurkanovic, line 499: rename,e.g. generate_step_equilibration_constraints (also rename the two above functions in the same style)

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+                        end
+                        eta_vec = [eta_vec;eta];
+                        obj.eta_vec = eta_vec;
+                        delta_h = obj.w.h(ii,jj) - obj.w.h(ii,jj-1);
+                        homotopy_eq = [eta*delta_h - sigma;eta*delta_h + sigma];
+                        obj.g.step_equilibration(ii,jj) = {homotopy_eq, [-inf;0], [0;inf]};
+                    end
+                end
+                obj.eta_fun = Function('eta_fun', {obj.w.w}, {eta_vec});
+              case 'mlcp'
```
nurkanovic, line 617: mlcp is what? Please use a more instructive name.

nurkanovic: i guess it is the linear case? 
at some point if it turns out that the linear is always better than the old nonlinear ones, we should just delete the old nonlinear (direct, direct homotopy...)

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+
+        function create_solver(obj, solver_options, plugin)
+            if ~obj.populated
+                obj.populate_problem()
+            end
+
+            if ~exist('plugin')
+                plugin = 'scholtes_ineq';
+            end
+            obj.w.sort_by_index();
```
nurkanovic, line 686: this is to get the block banded sparsity right? if yes so, please add a comment here.

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+        end
+
+        function results = get_results_struct(obj)
+            opts = obj.opts;
+            model = obj.model;
+
+            rbp = ~opts.right_boundary_point_explicit;
+            
+            if opts.right_boundary_point_explicit
+                results.x = obj.discrete_problem.w.x(:,:,obj.opts.n_s).res;
```
nurkanovic, line 724: discrete  -> discrete_time to be more specific

+nosnoc/+ocp/solver.m
```diff
@@ -0,0 +1,85 @@
+classdef solver < handle
+    properties % TODO separate these by Get/Set access
+        model
+        opts
+        solver_opts
+
+        dcs
+        discrete_problem
```
nurkanovic, line 8: discrete_time_problem

+nosnoc/Options.m
```diff
@@ -0,0 +1,295 @@
...
+        lift_complementarities(1,1) logical = 0
+        experimental_supervertical_form(1,1) logical = 0
+        lower_bound_comp_lift(1,1) logical = 0
+
+        % TODO: Initialization is really a solver concern, but there is not really a good extraction approach yet.
+        % Initialization - Stewart
+        lp_initialization(1,1) logical = 0
+
+        % Initialization - Step
+        initial_alpha(1,1) double {mustBeReal, mustBeFinite} = 1
```
nurkanovic, line 65: It is worth thinking whether these values live in a given model (and passeed to the dcs latter) instead of the options?

+nosnoc/Options.m
```diff
@@ -0,0 +1,295 @@
...
+
+        % Relaxation of terminal (or stage for equidistant grids) numerical/phyisical time constraints
+        relax_terminal_numerical_time(1,1) logical = 0; % instead of imposing \sum_h = T, add it as ell1_penalty term
+        rho_terminal_numerical_time(1,1) double {mustBeNonnegative} = 1e2
+        relax_terminal_numerical_time_homotopy (1,1) logical = 0; % us the homotopy parameter for the penalty
+        relax_terminal_physical_time(1,1) logical = 0; % instead of imposing \sum_h = T, add it as ell1_penalty term
+        rho_terminal_physical_time(1,1) double {mustBeNonnegative} = 1e2
+        relax_terminal_physical_time_homotopy (1,1) logical = 0; % us the homotopy parameter for the penalty
+
+        cost_integration(1,1) logical = 1
```
nurkanovic, line 142: what happens if this is false? does it make sense to have it

+nosnoc/Options.m
```diff
@@ -0,0 +1,295 @@
...
+        % Experimental:
+        no_initial_impacts(1,1) logical = 0
+        
+
+        % All MPCC parameters
+        T_val(1,1) double {mustBePositive} = 1
+        p_val
+
+        % Butcher Tableu
+        A_irk double
```
nurkanovic, line 153: rename irk to rk?

+nosnoc/+dcs/stewart.m
apozharski: Yep, this is fixed.

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
+classdef stewart < vdx.problems.Mpcc
```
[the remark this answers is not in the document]
apozharski: It is clear from the fully qualified name `nosnoc.dcs.stewart` vs `nosnoc.discrete_time_problem.stewart` though I agree if using an editor that does not automatically suffix with the parent directory for files with the same name this can get annoying. I don't have a much better naming scheme however.

+nosnoc/+discrete_problem/stewart.m
apozharski: Agreed, and renamed.

+nosnoc/+discrete_problem/stewart.m
apozharski: Agreed, renamed.

+nosnoc/Options.m
apozharski: This settings' name is horrible, We use it when we want to write an estimator so we don't integrate the cost but only take it at the discrete sample (stage) points. I am open to suggestions.

+nosnoc/+discrete_problem/stewart.m
apozharski: I agree that it is ugly, however I have done a lot to improve performance of variable creation (to the point where it is 10x faster about for big problems to what we had before) in VDX which relies on this one-shot variable creation. I have a few ideas about how to maintain this speedup even with different numbers of finite elements but I believe this will take some backend work in VDX. 

I am updating to allow for variable `N_finite_elements` but adding a big todo comment to remind myself to do the backend work.

+nosnoc/+discrete_problem/stewart.m
apozharski: Agreed. Also we need to think about what namespace the Enums should live in/should we keep them in the global namespace due to matlab's horrible autocomplete.

+nosnoc/+discrete_problem/stewart.m
apozharski: Would `linear` be a better name? I think in some sense `lcp`/`mlcp` is less deceptive as it explicitly points out that it formulates step-eq as a complementarity problem.

+nosnoc/+discrete_problem/stewart.m
```diff
@@ -0,0 +1,771 @@
...
+            opts = obj.opts;
+            v_global = obj.w.v_global();
+            p_global = obj.p.p_global();
+
+            if ~opts.use_fesd % do nothing
+                return
+            end
+
+            switch obj.opts.step_equilibration
+              case 'heuristic_mean'
```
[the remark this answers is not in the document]
apozharski: Yes it is.

+nosnoc/Options.m
nurkanovic: Okay, it is fine, just adding the very same comment to the setting explaining it is sufficient.

nurkanovic:
The new structure is very clean and well-organized. Before we merge, I have a few comments to polish it further:

- Maybe the files in different folders should not have the same name? 
E.g. we have several files Base, Stewart, etc.


- Licence header missing in new functions

- ocp_solver methods should have the same name formatting, e.g. get_x, get_u (instead of getU).

- properties in an object that are not used should be removed? e.g. model.lsq_T, model.g_path should be removed if they are empty (I see that they are used to define "empty" path constraints, but it would be nice to have the output as light as possible). The corresponding flags are anyway created in model\Base

- Similarly, having e.g. x_ref_val = zeros(...,...) is somewhat confusing, having them not at all in the output if not defined by the user would be nice.

- rename ocp_solver.opts  -> ocp_solver.problem_options , ocp_solver.solver_opts  -> ocp_solver.solver_options

- why are  g_Stewart: [], sys_idx: [] always empty in .dcs Stewart?

- replace h_res = ocp_solver.discrete_time_problem.w.h.res; by more readable syntax like h_res = ocp_solver.get('h'); other examples are:
 theta = ocp_solver.get_full('theta'), If a variable has a lower depth, then get and get_full should give the same resuls, e.g. h_res = ocp_solver.get_full('h') is same as h_res = ocp_solver.get('h');

- there are some capitalizations to be discussed, e.g. nosnoc.ocp.solver(...), nosnoc.solver.options() look nicer then the last letter capitalized.
   problem_options = nosnoc.Options() or problem_options = nosnoc.options()?

- In nosnoc, model, Base:         % TODO: maybe a separate OCP class common to all model types? - please elaborate, why more classes?

- g_comp_path -> how it is currently passed? as a n_path_comp x 2 matrix? How is then g_comp_path_fun handled? 
Should we maybe have a G_path and H_path instead, to me more consistent with our storing of complementarity constraints elsewhere?

- the name g_switching and g_switching_fun are a bit confusing (I know that i coined them), but something more instructive could be used, e.g. g_lp_lagrangian
  or g_lp_stationarity

apozharski:
A few responses here:
>     * Maybe the files in different folders should not have the same name?
>       E.g. we have several files Base, Stewart, etc.
My opinion is this is fine.  I prefer the clean-ness of `model.Stewart` (or even `model.stewart` as you suggest later) to something like `model.StewartModel`. I am not 100% set on this and if it causes development issues with the mainline matlab IDE then we can change this. it does not cause trouble with my emacs setup as it automatically tags each file with enough of its parent directories to make it unique, i.e. `Stewart.m<+model>`  vs `Stewart.m<+dcs>`.

>     * Licence header missing in new functions
I think this is something to be tackled in the near future (pre v1.0?). I will write a quick script to automatically update the BSD licence header in the required `.m` files to include the current year and contain all the developers etc. 

>     * ocp_solver methods should have the same name formatting, e.g. get_x, get_u (instead of getU).
Agreed

>     * properties in an object that are not used should be removed? e.g. model.lsq_T, model.g_path should be removed if they are empty (I see that they are used to define "empty" path constraints, but it would be nice to have the output as light as possible). The corresponding flags are anyway created in model\Base
>     * Similarly, having e.g. x_ref_val = zeros(...,...) is somewhat confusing, having them not at all in the output if not defined by the user would be nice.
What do you mean by "removed". There are two ways to do that for objects of classes: Actually removing via [DynamicProps](https://www.mathworks.com/help/matlab/matlab_oop/dynamic-properties-adding-properties-to-an-instance.html) or just modifying the display to "hide" the unused properties via [custom display functions](https://www.mathworks.com/help/matlab/group_analysisopt.html). I use both to accomplish the results you see when you display `vdx.Vector` subclasses with dummy dynamic properties and some other custom work to display the relevant properties. This can absolutely be done but could you clarify what you mean.  

>     * rename ocp_solver.opts  -> ocp_solver.problem_options , ocp_solver.solver_opts  -> ocp_solver.solver_options
I really prefer opts to options, but I will concede the point here.  

>     * why are  g_Stewart: [], sys_idx: [] always empty in .dcs Stewart?
Because we use `model.Stewart.g_ind` for the first one and, sys_idx was something I thought I needed but is now vestigial. Will remove.

>     * replace h_res = ocp_solver.discrete_time_problem.w.h.res; by more readable syntax like h_res = ocp_solver.get('h'); other examples are:
>       theta = ocp_solver.get_full('theta'), If a variable has a lower depth, then get and get_full should give the same resuls, e.g. h_res = ocp_solver.get_full('h') is same as h_res = ocp_solver.get('h');
No disagreement here.

>     * there are some capitalizations to be discussed, e.g. nosnoc.ocp.solver(...), nosnoc.solver.options() look nicer then the last letter capitalized.
>       problem_options = nosnoc.Options() or problem_options = nosnoc.options()?


>     * In nosnoc, model, Base:         % TODO: maybe a separate OCP class common to all model types? - please elaborate, why more classes?
The idea here was to separate model definition from OCP definition, i,e. model only contains things necessary for dynamical definition of the system and an ocp class contains the necessary things for an OCP. I believe this idea was discussed and rejected previously.

>     * g_comp_path -> how it is currently passed? as a n_path_comp x 2 matrix? How is then g_comp_path_fun handled?
>       Should we maybe have a G_path and H_path instead, to me more consistent with our storing of complementarity constraints elsewhere?
Yes currently it is passed that way. I have no strong opinions on storing it as `G_path` and `H_path`. Will re-implement that way.

>     * the name g_switching and g_switching_fun are a bit confusing (I know that i coined them), but something more instructive could be used, e.g. g_lp_lagrangian
>       or g_lp_stationarity

`g_lp_stationarity` is fine with me I think.

apozharski:
image
Current custom display implementation example.

nurkanovic:
My previous comments where addressed in a satisfactory way. I looked now in detail through all new files, and we are close to merge the PR. 
I added couple of comments in the code to help refresh ones memory when looking at it with some time distance.

Before we finish, I have few rather small things left:

1) integrator.simulate() should have the same verbose as the previous integrator had (e.g. how much in time it advanced) - it was useful for debugging.

2) it is a bit dissatisfying that integrator.stats stores the stats only of the last integration step. sometimes it is useful to plot the cpu time of each integration step (this can be a separate issue)

3) the integrator does not have anymore what used to be result.extended.t_grid, i.e. the time grid that has the collocation time points in between. this is useful for visualization, and debugging.

4) it seems redundant to have both ocp_solver.get_x() and ocp_solver.get('x') if they get the same result. I would keep only the latter one.	

5) What is the purpose of ocp_solver.getResults()  (get_results() would be a more consistent name)

6) ocp_solver.dcs should also print x (and possibly u) in the variables. 

7)   % time varying parameters
        p_time_var
        p_time_var_val
        % params
        p_time_var_stages
        % all params
        p
	Please comment in the code the differences. Not clear why p_time_var_stages has no _val and what is the difference to p_time_var
	
8) It would be useful to have also upper bounds on G_path and H_path (but can be a different small Issue/PR)

9) Step equilibration in discrete_time_problem/Stewart , obj.w.lambda_theta(ii,2:opts.N_finite_elements(ii)) = {{'lambda_theta', dims.n_theta},0,inf};  obj.w.lambda_lambda(ii,2:opts.N_finite_elements(ii)) = {{'lambda_lambda', dims.n_lambda},0,inf}; the variable names are not instructive.

10) Please justify why x_0 = obj.w.x(0,0,opts.n_s) and not x_0 = obj.w.x(0,0,1) or x_0 = obj.w.x(0,0,0) (also while defining it)

11) StepEquilibrationMode.mlcp might have a more instructive/explicit name

apozharski:
> My previous comments where addressed in a satisfactory way. I looked now in detail through all new files, and we are close to merge the PR. I added couple of comments in the code to help refresh ones memory when looking at it with some time distance.
> 
> Before we finish, I have few rather small things left:
> 
>     1. integrator.simulate() should have the same verbose as the previous integrator had (e.g. how much in time it advanced) - it was useful for debugging.
Addressed
>     2. it is a bit dissatisfying that integrator.stats stores the stats only of the last integration step. sometimes it is useful to plot the cpu time of each integration step (this can be a separate issue)
Addressed
>     3. the integrator does not have anymore what used to be result.extended.t_grid, i.e. the time grid that has the collocation time points in between. this is useful for visualization, and debugging.
Addressed
>     4. it seems redundant to have both ocp_solver.get_x() and ocp_solver.get('x') if they get the same result. I would keep only the latter one.
Addressed
>     5. What is the purpose of ocp_solver.getResults()  (get_results() would be a more consistent name)
Removed, there due to old design idea.
>     6. ocp_solver.dcs should also print x (and possibly u) in the variables.
Added. I think all the display code will end up being a bit of a chore to update but that is the life of providing a nice user experience.
>     7. % time varying parameters
>        p_time_var
>        p_time_var_val
>        % params
>        p_time_var_stages
>        % all params
>        p
>        Please comment in the code the differences. Not clear why p_time_var_stages has no _val and what is the difference to p_time_var
Frankly `p_time_var_stages` is never used anywhere except as an excuse to generate all of the symbolics at model time. I am not sure why this would be useful unless you want to for some reason have time varying params affect other time intervals which would be a horrible hack. As such this was removed.

>     8. It would be useful to have also upper bounds on G_path and H_path (but can be a different small Issue/PR)
Agree but imo this is a minor thing and should indeed be a separate issue.

>     9. Step equilibration in discrete_time_problem/Stewart , obj.w.lambda_theta(ii,2:opts.N_finite_elements(ii)) = {{'lambda_theta', dims.n_theta},0,inf};  obj.w.lambda_lambda(ii,2:opts.N_finite_elements(ii)) = {{'lambda_lambda', dims.n_lambda},0,inf}; the variable names are not instructive.
I don't have better instructive names that aren't significantly longer. And while we are no longer limited to 80 columns by punch cards I still think a shorter name that is accurate (`lambda_lambda` is the multiplier for the `lambda` KKT conditions) is far more readable than something longer.

>     10. Please justify why x_0 = obj.w.x(0,0,opts.n_s) and not x_0 = obj.w.x(0,0,1) or x_0 = obj.w.x(0,0,0) (also while defining it)
There are several (one major) ergonomic reasons to do this, as well as a philosophical one:
It allows clean extraction from  `vdx.Variable` as it lets you do `problem.w.var(:,:,opts.n_s).res` which gets the full trajectory including the initial point. 
Further on a philosophical level the initial position _is_ the end point of some previous process.

To me the ergonomic reasons are sufficient to overrule basically anything as making the _extremely_ common task of data extraction take extra lines for no reason simply increases the static friction one must overcome when extending the code.
 
>     11. StepEquilibrationMode.mlcp might have a more instructive/explicit name
We could perhaps call it `linear` or `linear_complementarity`. Though those really don't give away much more about what is going on.

nurkanovic (approved):
All comments are addressed - ready to be merged! GJ.

PR #99, opened 2024-05-16
New design Heaviside pipeline

apozharski:


nurkanovic:
There are a few minor bugs I found so far:
- In new_design_car_time_opt.m  - the model class created as Heaviside does not recognize S as a field. 
- one must use problem_options.dcs_mode = "Step", since "Heaviside" is not recognized, this should be fixed.
- new_design_friction_blocks_stewart - simulating with Heaviside, one of the subproblem becomes infeasible for some reason - is it again maybe a bad initialization for the complementarity variables? (one may also look did it converge with the old implementation, just to isolate bugs)

+nosnoc/+model/Heaviside.m
```diff
@@ -0,0 +1,46 @@
+classdef Heaviside < nosnoc.model.Base
+    properties
+        f_x
+
+        c
```
nurkanovic, line 5: this already exists in base, why here again?

+nosnoc/+model/Heaviside.m
```diff
@@ -0,0 +1,46 @@
+classdef Heaviside < nosnoc.model.Base
+    properties
```
nurkanovic, line 2: Add short comment  (1-2 sentences) why is this not a PSS in some cases (will be in docu anyway)

+nosnoc/+dcs/Heaviside.m
```diff
@@ -0,0 +1,229 @@
...
+                obj.lambda_p_sys{ii} = define_casadi_symbolic(opts.casadi_symbolic_mode,['lambda_p_' ii_str],obj.dims.n_c_sys(ii));
+                obj.lambda_p = [obj.lambda_p;obj.lambda_p_sys{ii}];
+            end
+
+            % symbolic variables z = [alpha;obj.lambda_n;obj.lambda_p];
+            obj.z_all = [obj.alpha;obj.lambda_n;obj.lambda_p;obj.model.z];
+            obj.dims = dims;
+        end
+
+        function generate_variables_from_heaviside(obj, opts)
```
nurkanovic, line 90: the functions _pss and _heaviside (later also for equaitons)  share a lot of code (Except for the theta and f_x parts, they are almost indentical). 
What speak against of having one function for vars and eqs, and have a if-else or switch-case?

nurkanovic:
> ###
> **[nurkanovic](/nurkanovic) ** commented [Jun 18, 2024](#issuecomment-2176332967)

I am not sure yet how to resolve, if one wants to use PSS with the Heaviside reformulation, but calls the Heaviside model by creating the model object. 
One way it to support both options if it does not make code to complicated, or to just throw an instructive error (e.g. when trying to put in an S into an Heaviside).

apozharski:
> > **[nurkanovic](/nurkanovic) ** commented [Jun 18, 2024](#issuecomment-2176332967)
> 
> I am not sure yet how to resolve, if one wants to use PSS with the Heaviside reformulation, but calls the Heaviside model by creating the model object. One way it to support both options if it does not make code to complicated, or to just throw an instructive error (e.g. when trying to put in an S into an Heaviside).

My view on this is RTFM :sweat_smile:. If a user uses the wrong class without reading even the top level documentation and gets `Unrecognized property 'S' for class 'nosnoc.model.Heaviside'` and doesn't immediately go to the docs I don't think that user can be saved.

+nosnoc/+model/Heaviside.m
apozharski: I don't believe it is in `nosnoc.model.Base`. As a matter of fact matlab will loudly complain if you double define a property.

+nosnoc/+dcs/Heaviside.m
apozharski: No extremely strong opinions here. I think in this case we can absolutely have an if statement. The only thing is sometimes it is nice to have a pseudo-linear way to read the code even if it requires some repetition. Perhaps this is an over-correction on my part however, as I have been quite frustrated during this rewrite with the interleaving `if`s in the old design and disentangling them one by one.

apozharski:
> new_design_friction_blocks_stewart - simulating with Heaviside, one of the subproblem becomes infeasible for some reason - is it again maybe a bad initialization for the complementarity variables? (one may also look did it converge with the old implementation, just to isolate bugs)

This is not a bug per se, but it is the fact that `cross_comp_mode=3` and `cross_comp_mode=4` were flipped. I need to check the Stewart version away.

+nosnoc/+dcs/Heaviside.m
nurkanovic: I agree, but 90% is repetition and adds quite a few extra lines. Well, most of the old ifs are mostly resolved by having extra classes for every problem type. 

I can live with the repetition, but shorter and still understandable code is also not bad.

apozharski:
> What confused even me, that the commented code in the examples was suggesting that this is a valid path. :|

Hmm well that is definitely something we should avoid then. Strictly speaking the point is that the `Heaviside` DCS reformulation should take a PSS _or_ a Heaviside model, and theoretically one _could_ directly formulate  a PSS as an equivalent Heaviside model but that takes some thought.

nurkanovic (approved):
LGTM GJ

Issue #96, opened 2024-05-15
Rleaxed and slacked constraints in nosnoc


nurkanovic:
Currently, we support ell_1 relaxation of some terminal constraints
1) the terminal constraints of the ocp
2) some of the FESD/time-freezing terminal constraints to meet a certain final numerical/physical time. 

I would still steer the relaxation of these constraints with appropriate separate options, as we have them now.

In addition to this, it would be nice to be able to relax other constraints (e.g. using vdx internally to define according variables and constraints). In particular it would be nice to relax:
1) Path constraints
2) Dynamics constraints (maybe to separate even between the differential and algebraic part of the dynamics)

Choices for relaxing:
1) ell 2 (every constraint gets a separate slack), maybe we can have upper bounds on slacks to avoid unbounded problems
2) ell1 (every constraint gets a separate slack)
3) ell_inf (every constraint group gets one slack)

Choice for penalty parameter:
1) fixed value (separate parameter for every constraint group, by groups i mean: path, dynamics, terminal, terminal_algoritmic (nice and instructive names are a plus))
2) use rho = 1/sigma - we should decide if we can mix fixed values and homotopy penalities. one one hand it gives a lot of flexibility, on the other hand too many choices might be confusing.

nurkanovic:
We should have minimal examples and sanity checks for every of the mode.

PR #100, opened 2024-05-17
New design  PDS/FOSwP pipeline

apozharski:


examples/new_design_examples/new_design_pds.m
nurkanovic, line 1: There are many intermediate integration steps infeasible, or maximum iterations reached with the current settings.

examples/new_design_examples/new_design_pds.m
```diff
@@ -0,0 +1,52 @@
...
+
+model.x0 = [0;pi-2];
+
+integrator = nosnoc.Integrator(model, problem_options, solver_options);
+[t_grid, x_res] = integrator.simulate();
+%
+figure
+plot(x_res(1,:), x_res(2,:))
+grid on
+xlabel('$t$','Interpreter','latex')
```
nurkanovic, line 45: the labels are x_1 and x_2, not t, and x

examples/new_design_examples/plot_moving_set.m
nurkanovic, line 1: cool example, but when the outer circle sweeps the inner, they are not quite in contact.
Is this a plotting artefact?

examples/new_design_examples/new_design_pds.m
```diff
@@ -0,0 +1,52 @@
...
+
+model.x0 = [sqrt(2)/2;sqrt(2)/2];
+x = SX.sym('x',2);
+model.x = x;
+model.c = x(2)+0.2;
+model.f_x = [x(2);-x(1)];
+
+model.x0 = [0;pi-2];
+
+integrator = nosnoc.Integrator(model, problem_options, solver_options);
```
nurkanovic, line 39: I find it hard to change the mpec solver currently... ideally, we pass it through the solver options.

+nosnoc/+model/Pds.m
```diff
@@ -0,0 +1,43 @@
+classdef Pds < nosnoc.model.Base
+    properties
+        f_x
+        c
+        E
```
nurkanovic, line 5: Is this matrix already used and mature, or is it unpublished research?

examples/new_design_examples/new_design_moving_set.m
```diff
@@ -0,0 +1,46 @@
...
+model.f_x = [0; 0; 1];
+model.E = diag([1,1,0]);
+
+integrator = nosnoc.Integrator(model, problem_options, solver_options);
+[t_grid, x_res] = integrator.simulate();
+%
+figure
+plot(x_res(1,:), x_res(2,:))
+grid on
+xlabel('$t$','Interpreter','latex')
```
nurkanovic, line 41: $x_1$

examples/new_design_examples/new_design_moving_set.m
```diff
@@ -0,0 +1,46 @@
...
+ylabel('$x(t)$','Interpreter','latex')
```
nurkanovic, line 42: $x_2$

examples/new_design_examples/plot_moving_set.m
nurkanovic, line 1: I would put already all PDS examples in respective folders into the main examples folder. 

Also, it is time to translate the Stewart and Heaviside examples into the new design (and double-check do we still get exactly the same results)  - but this is a separate Issue/PR (I am willing to do it, as i want to redo our comments in the examples)

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
+classdef Gcs < nosnoc.dcs.Base
+    properties
+        lambda
+
+        f_x
+
+        dims
+
+        f_fun
+        nabla_c_fun
```
nurkanovic, line 10: add some comments what is what (maybe it is already compatible with our new read the docs)?

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+            import casadi.*
+            model = obj.model;
+            dims = obj.dims;
+
+            nabla_c = model.c.jacobian(model.x)';
+            
+            obj.f_x = model.f_x + model.E*nabla_c*obj.lambda;
+
+            obj.f_x_fun = Function('f_x', {model.x, model.z, obj.lambda, model.u, model.v_global, model.p}, {obj.f_x, model.f_q});
+            obj.f_fun = Function('f', {model.x, model.z, model.u, model.v_global, model.p}, {model.f_x});
```
nurkanovic, line 41: maybe a bit more instructive name than f (which we sometimes use for the objective elsewhere)
f_uncstr (unconstrained), f_rhs, f_ode are some examples

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+        function generate_equations(obj, opts)
+            import casadi.*
+            model = obj.model;
+            dims = obj.dims;
+
+            nabla_c = model.c.jacobian(model.x)';
+            
+            obj.f_x = model.f_x + model.E*nabla_c*obj.lambda;
+
+            obj.f_x_fun = Function('f_x', {model.x, model.z, obj.lambda, model.u, model.v_global, model.p}, {obj.f_x, model.f_q});
```
nurkanovic, line 40: if you have f_q_fun below, is it needed that you have f_q as output. 
I guess the first option is when you want to have a accurate integration of the objective (default), the second is for special cases like state estimation, right? (a one line comment in the code would be helpful for our future self)

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+            obj.f_fun = Function('f', {model.x, model.z, model.u, model.v_global, model.p}, {model.f_x});
+            obj.f_q_fun = Function('f_q', {model.x, model.z, obj.lambda, model.u, model.v_global, model.p}, {model.f_q});
+            obj.g_z_fun = Function('g_z', {model.x, model.z, model.u, model.v_global, model.p}, {model.g_z});
+            obj.g_alg_fun = Function('g_alg', {model.x, model.z, obj.lambda, model.u, model.v_global, model.p}, {[]});
```
nurkanovic, line 44: why is this empty output function defined? (used for latter developments like sdf) 

also comments what is in g_z and what in g_alg stored would be of great help.

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+            obj.c_fun = Function('c_fun', {model.x, model.z, model.v_global, model.p}, {model.c});
+            obj.nabla_c_fun = Function('c_fun', {model.x, model.z, model.v_global, model.p}, {nabla_c});
+            obj.g_path_fun = Function('g_path', {model.x, model.z, model.u, model.v_global, model.p}, {model.g_path}); % TODO(@anton) do dependence checking for spliting the path constriants
```
nurkanovic, line 47: regarding the todo, split in what? (e.g. if they depend only on control, but you have path on all rk stages, to avoid duplicating constraints, or is it something else)

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+            obj.f_lsq_T_fun = Function('f_lsq_T_fun',{model.x,model.x_ref_end,model.p_global},{model.f_lsq_T});
+        end
+    end
+
+    methods(Access=protected)
+        function propgrp = getPropertyGroups(obj)
+            propgrp = getPropertyGroups@nosnoc.dcs.Base(obj);
+            group_title = 'Variables';
+            var_list = struct;
+            var_list.lambda;
```
nurkanovic, line 63: not sure i understand this part, why only lambda for what? (not questioning the correctness, just want to understand)

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
+classdef Gcs < nosnoc.dcs.Base
+    properties
+        lambda
+
+        f_x
+
+        dims
+
+        f_fun
```
nurkanovic, line 9: i find it like on other similar occasion unclear, 
f_x is  f_unconstrained + nabla_c'*lambda;
and f_fun is made of f_unconstrained? more instructive names (and possibly comments would make it more sustainable)

+nosnoc/+dcs/Gcs.m
```diff
@@ -0,0 +1,67 @@
...
+        nabla_c_fun
+        c_fun
+        g_comp_path_fun
```
nurkanovic, line 12: why not in base? is this not the user path complementarity constrains? 
If not, what is it? Use more instructive name if not the above?

+nosnoc/+dcs/Gcs.m
nurkanovic, line 40: obj.f_x and model.f_x are not the same hence they may have different names to avoid ambiguity

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+        end
+
+        function create_variables(obj)
+            dims = obj.dcs.dims;
+            dcs = obj.dcs;
+            model = obj.model;
+            opts = obj.opts;
+
+            % Parameters
+            obj.p.rho_h_p = {{'rho_h_p',1}, 1};
```
nurkanovic, line 25: I guess we dont want a base discrete time problem, where all the parameters e.g., that the same live for sake of readiblity?

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            obj.p.T = {{'T',1}, opts.T};
+            obj.p.p_global = {model.p_global, model.p_global_val};
+
+
+            % 0d vars: Variables which only exist once globally. 
+            obj.w.v_global = {{'v_global',dims.n_v_global}, model.lbv_global, model.ubv_global, model.v0_global};
+            if opts.use_speed_of_time_variables && ~opts.local_speed_of_time_variable
+                obj.w.sot = {{'sot', 1}, opts.s_sot_min, opts.s_sot_max, opts.s_sot0};
+            end
+            if opts.time_optimal_problem
```
nurkanovic, line 36: please add a time optimal pds example, just to verify the correctness of this (e.g. you can use the standard simulation pds, add some control in 2d, and force that the "particle" ends somewhere on the constraint. for instance the control could change the vector field such that it does not leave the "sliding mode" but stays on it

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                else
+                    obj.w.x(ii,1:opts.N_finite_elements(ii),opts.n_s) = {{'x', dims.n_x}, model.lbx, model.ubx, model.x0};
+                end
+                if (opts.rk_representation == RKRepresentation.differential ||...
+                    opts.rk_representation == RKRepresentation.differential_lift_x)
+                    obj.w.v(ii,1:opts.N_finite_elements(ii),1:opts.n_s) = {{'v', dims.n_x}};
+                end
+                
+                obj.w.z(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
+                obj.w.lambda(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'lambda', dims.n_lambda},0, inf};
```
nurkanovic, line 103: what is the default value of lambda? should it be defined in the dcs.Gcs? I guess having an initialization for the complementarity variables not at (0,0) is helpful like in the othe cases.

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    obj.w.x(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'x', dims.n_x}, model.lbx, model.ubx, model.x0};
+                else
+                    obj.w.x(ii,1:opts.N_finite_elements(ii),opts.n_s) = {{'x', dims.n_x}, model.lbx, model.ubx, model.x0};
+                end
+                if (opts.rk_representation == RKRepresentation.differential ||...
+                    opts.rk_representation == RKRepresentation.differential_lift_x)
+                    obj.w.v(ii,1:opts.N_finite_elements(ii),1:opts.n_s) = {{'v', dims.n_x}};
+                end
+                
+                obj.w.z(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
```
nurkanovic, line 102: If the function c is nonlinear, then it makes sense to lift it (as we know comp. constraints dont like if not lifted)
how is this happening currently, at the mpcc level? or is somewhere a algebraic variable z_c (or whatever name) defined and an algebaric constraint to take care of it (i feel like i missed it in dcs.GCS, but is to me maybe acontinious time think for lifting  - for lifting theta and alpha, in Heaviside we also do it all before discretization)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                        % if only time_rescaling is true, speed of time and step size all lumped together, e.g., \hat{h}_{k,i} = s_n * h_{k,i}, hence the bounds need to be extended.
+                        ubh = (1+opts.gamma_h)*h0*opts.s_sot_max;
+                        lbh = (1-opts.gamma_h)*h0/opts.s_sot_min;
+                    end
+                    obj.w.h(ii,1:opts.N_finite_elements(ii)) = {{'h', 1}, lbh, ubh, h0};
+                end
+                if obj.opts.step_equilibration == StepEquilibrationMode.linear_complementarity
+                    obj.w.B_max(ii,2:opts.N_finite_elements(ii)) = {{'B_max', dims.n_lambda},-inf,inf};
+                    obj.w.pi_c(ii,2:opts.N_finite_elements(ii)) = {{'pi_c', dims.n_c},-inf,inf};
+                    obj.w.pi_lambda(ii,2:opts.N_finite_elements(ii)) = {{'pi_lambda', dims.n_lambda},-inf,inf};
```
nurkanovic, line 70: pi_lambda, 
c_lambda... it is very easy to forget what is what. can you please add some comments for our future self :D

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    ubh = (1 + opts.gamma_h) * h0;
+                    lbh = (1 - opts.gamma_h) * h0;
+                    if opts.time_rescaling && ~opts.use_speed_of_time_variables
+                        % if only time_rescaling is true, speed of time and step size all lumped together, e.g., \hat{h}_{k,i} = s_n * h_{k,i}, hence the bounds need to be extended.
+                        ubh = (1+opts.gamma_h)*h0*opts.s_sot_max;
+                        lbh = (1-opts.gamma_h)*h0/opts.s_sot_min;
+                    end
+                    obj.w.h(ii,1:opts.N_finite_elements(ii)) = {{'h', 1}, lbh, ubh, h0};
+                end
+                if obj.opts.step_equilibration == StepEquilibrationMode.linear_complementarity
```
nurkanovic, line 67: except the huristic mean, this is the only one supported in PDS, which is for me more than fine. 
However, if one picks some of the other options, we should switch to one of those, and throw a warning.

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                if obj.opts.use_fesd
+                    ubh = (1 + opts.gamma_h) * h0;
+                    lbh = (1 - opts.gamma_h) * h0;
+                    if opts.time_rescaling && ~opts.use_speed_of_time_variables
+                        % if only time_rescaling is true, speed of time and step size all lumped together, e.g., \hat{h}_{k,i} = s_n * h_{k,i}, hence the bounds need to be extended.
+                        ubh = (1+opts.gamma_h)*h0*opts.s_sot_max;
+                        lbh = (1-opts.gamma_h)*h0/opts.s_sot_min;
+                    end
+                    obj.w.h(ii,1:opts.N_finite_elements(ii)) = {{'h', 1}, lbh, ubh, h0};
+                end
```
nurkanovic, line 66: if use_fesd = 0, in the msv example, i get the following error:
`
unrecognized function or variable 'p'.

Error in nosnoc.discrete_time_problem.Gcs/generate_complementarity_constraints (line 524)
                            c_ijk = dcs.c_fun(x_ijk,z_ijk,v_global,p);

Error in nosnoc.discrete_time_problem.Gcs/populate_problem (line 769)
            obj.generate_complementarity_constraints();

Error in nosnoc.ocp.Solver (line 53)

Error in new_design_msv (line 54)
ocp_solver = nosnoc.ocp.Solver(model, problem_options, solver_options);
`

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            end
+                            obj.G.cross_comp(ii,jj) = {vertcat(Gij{:})};
+                            obj.H.cross_comp(ii,jj) = {vertcat(Hij{:})};
+                            lambda_prev = obj.w.lambda(ii,jj,opts.n_s + rbp);
+                            x_prev = obj.w.x(ii,jj,opts.n_s + rbp);
+                            z_prev = obj.w.z(ii,jj,opts.n_s + rbp);
+                            c_prev = dcs.c_fun(x_prev,z_prev, v_global, p);
+                        end
+                    end
+                  case CrossCompMode.FE_FE
```
nurkanovic, line 491: in e.g. the msv example i get strange results with this cross comp mode, one might check is it bug free (maybe in simpler simulation problems)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            p_global = obj.p.p_global();
+
+            x_prev = obj.w.x(0,0,opts.n_s);
+            z_prev = obj.w.z(0,0,opts.n_s);
+            lambda_prev = obj.w.lambda(0,0,opts.n_s);
+            c_prev = dcs.c_fun(x_prev,z_prev, v_global, [p_global;obj.p.p_time_var(1)]);
+
+            if opts.use_fesd
+                switch opts.cross_comp_mode
+                  case CrossCompMode.STAGE_STAGE
```
nurkanovic, line 399: in e.g. the msv example i get strange results with this cross comp mode, one might check is it bug free (maybe in simpler simulation problems)

examples/new_design_examples/new_design_msv.m
```diff
@@ -0,0 +1,98 @@
...
+import vdx.*
+
+R = 3.5;
+%% Define projected system
+problem_options = nosnoc.Options();
+solver_options = nosnoc.solver.Options();
+%problem_options.rk_scheme = RKSchemes.GAUSS_LEGENDRE;
+%problem_options.rk_scheme = RKSchemes.LOBATTO_IIIC;
+problem_options.rk_scheme = RKSchemes.RADAU_IIA;
+%problem_options.rk_representation= RKRepresentation.differential_lift_x;
```
nurkanovic, line 13: this option leads to strage results in my case.

examples/new_design_examples/new_design_msv.m
```diff
@@ -0,0 +1,98 @@
...
+R = 3.5;
+%% Define projected system
+problem_options = nosnoc.Options();
+solver_options = nosnoc.solver.Options();
+%problem_options.rk_scheme = RKSchemes.GAUSS_LEGENDRE;
+%problem_options.rk_scheme = RKSchemes.LOBATTO_IIIC;
+problem_options.rk_scheme = RKSchemes.RADAU_IIA;
+%problem_options.rk_representation= RKRepresentation.differential_lift_x; 
+problem_options.rk_representation = RKRepresentation.integral;
+problem_options.cross_comp_mode = CrossCompMode.STAGE_STAGE;
```
nurkanovic, line 15: this option leads to strage results in my case.

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            obj.g.lambda_explicit(ii,jj) = {[lambda_ijk-c*lambda_cf_ijk]};
+                            obj.g.dynamics(ii,jj,opts.n_s+1) = {x_ijk - x_ij_end};
+                            obj.g.c_lb(ii,jj,opts.n_s+1) = {dcs.c_fun(x_ijk, z_ijk, v_global, p), 0, inf};
+                            obj.g.z(ii,jj,opts.n_s+1) = {dcs.g_z_fun(x_ijk, z_ijk, ui, v_global, p)};
+                        end
+                        if ~opts.g_path_at_stg && opts.g_path_at_fe
+                            obj.g.path(ii,jj) = {dcs.g_path_fun(x_ijk, z_ijk, ui, v_global, p), model.lbg_path, model.ubg_path};
+                        end
+                      case RKRepresentation.differential
+                        error("Differential representation without lifting is unsupported for gradient complementarity systems")
```
nurkanovic, line 229: instead of an error i would throw a warning and do differential_lift_x

examples/new_design_examples/new_design_msv.m
```diff
@@ -0,0 +1,98 @@
...
+% Solve
+ocp_solver = nosnoc.ocp.Solver(model, problem_options, solver_options);
+ocp_solver.solve();
+
+%% plot
+fontsize = 12;
+x_res = ocp_solver.get('x');
+u_res = ocp_solver.get('u');
+u_rep = kron(u_res, ones(1,problem_options.N_finite_elements(1)));
+T_final = ocp_solver.get('T_final');
```
nurkanovic, line 61: if it is not a time optimal problem, this will throw an error.
maybe we just return then just the model.T in this case - so the code is not broken. 
often when solving time optimal problems, one tries a few fixed T solutions to debug

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: this also does not allow Lobbato methods. 
however, is it possible to have them in lifted_x form?

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: okay, i see that the msv example is a time optimal one.

examples/new_design_examples/new_design_msv.m
```diff
@@ -0,0 +1,98 @@
+clear all
+close all
+import casadi.*
+import vdx.*
+
+R = 3.5;
+%% Define projected system
+problem_options = nosnoc.Options();
+solver_options = nosnoc.solver.Options();
+%problem_options.rk_scheme = RKSchemes.GAUSS_LEGENDRE;
```
nurkanovic, line 10: this option does not work properly. extremely bad convergence. I guess it is not implemented properly because of the rbp?
In case yes, throw a warning and put radau iia.

But in general it might not have be bad to have GL. The resulting "sliding modes" ODEs are not of very high index like in CLS problems, hence they have decent properties.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: or it should be defined in options. a name for the option would be something like lambda_pds_0 or lambda0_pds

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            end
+
+            % 1d vars: Variables that are defined per control stage
+            obj.w.u(1:opts.N_stages) = {{'u', dims.n_u}, model.lbu, model.ubu, model.u0};
+            obj.p.p_time_var(1:opts.N_stages) = {{'p_time_var', dims.n_p_time_var}, model.p_time_var_val};
+            if opts.use_speed_of_time_variables && opts.local_speed_of_time_variable
+                obj.w.sot(1:opts.N_stages) = {{'sot', 1}, opts.s_sot_min, opts.s_sot_max, opts.s_sot0};
+            end
+
+            % TODO(@anton) This _severely_ hurts performance over the vectorized assignment by doing N_stages vertcats of
```
nurkanovic, line 48: this is more like a vdx issues? wont change the code here?

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    obj.w.B_max(ii,2:opts.N_finite_elements(ii)) = {{'B_max', dims.n_lambda},-inf,inf};
+                    obj.w.pi_c(ii,2:opts.N_finite_elements(ii)) = {{'pi_c', dims.n_c},-inf,inf};
+                    obj.w.pi_lambda(ii,2:opts.N_finite_elements(ii)) = {{'pi_lambda', dims.n_lambda},-inf,inf};
+                    obj.w.lambda_c(ii,2:opts.N_finite_elements(ii)) = {{'lambda_c', dims.n_c},0,inf};
+                    obj.w.lambda_lambda(ii,2:opts.N_finite_elements(ii)) = {{'lambda_lambda', dims.n_lambda},0,inf};
+                    obj.w.eta(ii,2:opts.N_finite_elements(ii)) = {{'eta', dims.n_lambda},0,inf};
+                    obj.w.nu(ii,2:opts.N_finite_elements(ii)) = {{'nu', 1},0,inf};
+                end
+                if ~opts.right_boundary_point_explicit
+                    obj.w.not_c_primals(ii,1:opts.N_finite_elements(ii)) = {{'not_c_primal', 2}, 0 inf, 0.5};
```
nurkanovic, line 77: having for the default hard coded numbers that one cannot change easily is maybe not nice

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                end
+            end
+
+            % For c_n ~= 1 case
+            rbp = ~opts.right_boundary_point_explicit;
+            
+            % 3d vars: Variables defined on each rk stage
+            %          some of which are also defined at the initial point:
+            obj.w.x(0,0,opts.n_s) = {{['x_0'], dims.n_x}, model.x0, model.x0, model.x0};
+            obj.w.z(0,0,opts.n_s) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
```
nurkanovic, line 88: this are user algebraics? lifting algebraic, sdf algebarics? 
i always have to stop and think whenever i see z. adding comments in few occasions (also in other .m files than this would help our present and future self)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            obj.w.lambda(0,0,opts.n_s) = {{['lambda'], dims.n_lambda},0,inf};
+            for ii=1:opts.N_stages
+                if (opts.rk_representation == RKRepresentation.integral ||...
+                    opts.rk_representation == RKRepresentation.differential_lift_x)
+                    obj.w.x(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'x', dims.n_x}, model.lbx, model.ubx, model.x0};
+                else
+                    obj.w.x(ii,1:opts.N_finite_elements(ii),opts.n_s) = {{'x', dims.n_x}, model.lbx, model.ubx, model.x0};
+                end
+                if (opts.rk_representation == RKRepresentation.differential ||...
+                    opts.rk_representation == RKRepresentation.differential_lift_x)
```
nurkanovic, line 98: variables get defined for differential even tho it is not really supported.
also for people other than you and me, adding a comment what is v (a stage value for f_x might improve readibility)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            lambda_0 = obj.w.lambda(0,0,opts.n_s);
+            
+            obj.g.z(0,0,opts.n_s) = {dcs.g_z_fun(x_0, z_0, obj.w.u(1), v_global, [p_global;obj.p.p_time_var(1)])};
+            obj.g.c_lb(0,0,opts.n_s) = {dcs.c_fun(x_0, z_0, v_global, [p_global;obj.p.p_time_var(1)]), 0, inf};
+            
+            x_prev = obj.w.x(0,0,opts.n_s);
+            for ii=1:opts.N_stages
+                h0 = obj.p.T().val/(opts.N_stages*opts.N_finite_elements(ii));
+                
+                ui = obj.w.u(ii);
```
nurkanovic, line 141: u_i

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    switch opts.rk_representation
+                      case RKRepresentation.integral
+                        % In integral representation stage variables are states.
+                        x_ij_end = opts.D_rk(1)*x_prev;
+                        for kk=1:opts.n_s
+                            x_ijk = obj.w.x(ii,jj,kk);
+                            z_ijk = obj.w.z(ii,jj,kk);
+                            lambda_ijk = obj.w.lambda(ii,jj,kk);
+
+                            fj = s_sot*dcs.f_x_fun(x_ijk, z_ijk, lambda_ijk, ui, v_global, p);
```
nurkanovic, line 178: u_i ... and elsewhere

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            qj = s_sot*dcs.f_q_fun(x_ijk, z_ijk, lambda_ijk, ui, v_global, p);
+                            xk = opts.C_rk(1, kk+1) * x_prev;
+                            for rr=1:opts.n_s
+                                x_ijr = obj.w.x(ii,jj,rr);
+                                xk = xk + opts.C_rk(rr+1, kk+1) * x_ijr;
+                            end
+                            obj.g.dynamics(ii,jj,kk) = {h * fj - xk};
+                            obj.g.z(ii,jj,kk) = {dcs.g_z_fun(x_ijk, z_ijk, ui, v_global, p)};
+                            obj.g.c_lb(ii,jj,kk) = {dcs.c_fun(x_ijk, z_ijk, v_global, p), 0, inf};
```
nurkanovic, line 187: once again, i vote for a lifted version of $c$ if it is nonlinear (maybe to have it even as default setting - this would require some testing)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            if opts.cost_integration
+                                % also integrate the objective
+                                obj.f = obj.f + opts.B_rk(kk+1)*h*qj;
+                            end
+                        end
+                        if ~opts.right_boundary_point_explicit
+                            x_ijk = obj.w.x(ii,jj,opts.n_s+1);
+                            z_ijk = obj.w.z(ii,jj,opts.n_s+1);
+                            lambda_ijk = obj.w.lambda(ii,jj,opts.n_s+1);
+                            mu_not_c_ij = obj.w.mu_not_c(ii,jj);
```
nurkanovic, line 209: i am confused by the names _not_ 
please add some comments and elaborate

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            obj.g.lift_x(ii,jj,kk) = {x_ijk - x_temp};
+                        end
+                        x_ij_end = x_prev;
+                        for kk=1:opts.n_s
+                            x_ijk = obj.w.x(ii,jj,kk);
+                            v_ijk = obj.w.v(ii,jj,kk);
+                            z_ijk = obj.w.z(ii,jj,kk);
+                            lambda_ijk = obj.w.lambda(ii,jj,kk);
+                           
+                            fj = s_sot*dcs.f_x_fun(x_ijk, z_ijk, lambda_ijk, ui, v_global, p);
```
nurkanovic, line 249: for sake of consistency, f_j, q_j ? (and elsewhere)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    x_prev = obj.w.x(ii,jj,opts.n_s+rbp);
+                end
+                if ~opts.g_path_at_stg && ~opts.g_path_at_fe
+                    x_i = obj.w.x(ii, opts.N_finite_elements(ii), opts.n_s);
+                    z_i = obj.w.z(ii, opts.N_finite_elements(ii), opts.n_s);
+                    obj.g.path(ii) = {dcs.g_path_fun(x_i, z_i, ui, v_global, p), model.lbg_path, model.ubg_path};
+                end
+
+                % Least Squares Costs
+                % TODO we should convert the refs to params
```
nurkanovic, line 295: add issue for this todo

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    obj.f = obj.f + t_stage*dcs.f_lsq_u_fun(obj.w.u(ii),...
+                        model.u_ref_val(:,ii),...
+                        p);
+                end
+                if ~opts.cost_integration
+                    obj.f = obj.f + dcs.f_q_fun(x_ijk, z_ijk, lambda_ijk, ui, v_global, p);
+                end
+
+                % Clock Constraints
+                % TODO(@anton) HERE BE DRAGONS. This is by far the worst part of current nosnoc as it requires the discrete problem
```
nurkanovic, line 311: this comment here is irrelevant. we dont have time freezing for PDS (and first order systems in general, as they dont have state jumps do to constraint, in order to be meaningful)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                %              to understand something about the time-freezing reformulation which is ugly.
+                if obj.opts.use_fesd && opts.equidistant_control_grid
+                    if opts.time_optimal_problem
+                        if opts.use_speed_of_time_variables
+                            obj.g.equidistant_control_grid(ii) = {[sum_h - opts.h;sot*sum_h - obj.w.T_final()/opts.N_stages]};
+                        else
+                            obj.g.equidistant_control_grid(ii) = {sum_h - obj.w.T_final()/opts.N_stages};
+                        end
+                    else
+                        if opts.relax_terminal_numerical_time
```
nurkanovic, line 321: i love this

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                            % TODO(@armin) why is this allowed to be negative?
```
nurkanovic, line 322: relax, relax, relax  is the motto :D

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                    model.x_ref_end_val,...
+                    p_global);
+            end
+
+            % Terminal constraint
+            if opts.relax_terminal_constraint_homotopy
+                error("Currently unsupported")
+            end
+            g_terminal = dcs.g_terminal_fun(x_end, z_end, v_global, p_global);
+            switch opts.relax_terminal_constraint
```
nurkanovic, line 352: add TODO or even issue, 
to refactor this constrain relaxation as soon as the VDX relaxation is mature. 
But in general, i would still ask via options which group/type of constraints should be relaxed and how 
e.g. only time constraints, terminal constraints, path constraints, dynamics constraints (diff and algebaric) and have meaningful defaults. this flexiblity might be crucial for the really hard problems

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                        eta = 1;
+                        for jjj=1:length(nu)
+                            eta = eta*nu(jjj);
+                        end
+                        eta_vec = [eta_vec;eta];
+                        delta_h = obj.w.h(ii,jj) - obj.w.h(ii,jj-1);
+                        obj.f = obj.f + obj.p.rho_h_p() * eta * delta_h.^2
+                    end
+                end
+              case StepEquilibrationMode.direct
```
nurkanovic, line 630: I am not sure does this option make sense at all.
the cross comps are also complementarity constraints. and we should let them be treated by the mpcc solver.
the only things we distinguish are they 
* in the objective (then does the indicator function also influce the weight, or is jus the rho_h, second branching, is it mean or diff). 
* hard constraints

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+            if ~obj.populated
+                obj.populate_problem();
+            end
+
+            if ~exist('plugin')
+                plugin = 'scholtes_ineq';
+            end
+
+            % Sort by indices to recover almost block-band structure.
+            obj.w.sort_by_index();
```
nurkanovic, line 785: not related to this PR, but plotting the Jacobians of the constrains with  FESD ON and OFF would be interesting (also after introducing a clock state for the integration, to get rid of sum h constraints that interfere with a block band structure, cf. https://github.com/nurkanovic/nosnoc/issues/102)

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -0,0 +1,870 @@
...
+                results.mu = obj.discrete_time_problem.w.mu(:,:,obj.opts.n_s).res;
+                results.theta = obj.discrete_time_problem.w.theta(:,:,obj.opts.n_s).res;
+            else
+                results.x = [obj.discrete_time_problem.w.x(0,0,obj.opts.n_s).res,...
+                    obj.discrete_time_problem.w.x(1:opts.N_stages,:,obj.opts.n_s+1).res];
+                results.z = [obj.discrete_time_problem.w.z(0,0,obj.opts.n_s).res,...
+                    obj.discrete_time_problem.w.z(1:opts.N_stages,:,obj.opts.n_s+1).res];
+                results.lambda = [obj.discrete_time_problem.w.lambda(0,0,obj.opts.n_s).res,...
+                    obj.discrete_time_problem.w.lambda(1:opts.N_stages,:,obj.opts.n_s+1).res];
+                results.mu = [obj.discrete_time_problem.w.mu(0,0,obj.opts.n_s).res,...
```
nurkanovic, line 835: i am confused, why do you have a mu and theta in a GCS? they remind me of stewart dcs...
maybe pick different names + explain them

examples/new_design_examples/plot_moving_set.m
apozharski: Yes this is a plotting artifact, the center of the outer circle was calculated with a time that was one element behind hence the weird looking frames.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: This is a copy pate error :sweat_smile:. removed.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: I believe this is #96?

examples/new_design_examples/new_design_pds.m
apozharski: I did not realize the plugin option was missing from `Integrator.simulate`.  You can now change the MPEC solver like this:
```
[t_grid, x_res] = integrator.simulate('natural_residual_ineq');
```
for example.

apozharski: Though I think this needs to be discussed a bit more. Like I said before if we have non-relaxation solvers we plug in we should probably have that as the top level plugin passed to `nosnoc.solver.mpccsol` instead of the `nosnoc.solver.MpccMethod` as we do now. In the past I have hacked together interfaces for the solvers of Lin and De Marchi which could easily (I think) be transferred over to this framework relatively easily as a proof of concept for this.

+nosnoc/+model/Pds.m
apozharski: I guess this is unpublished research. It is  practically quite robust, however theory remains to be done.

examples/new_design_examples/new_design_pds.m
nurkanovic: I still believe, that the most intuitive way is pass it via solver_options. 
But keeping on top of this argument option next to it is also good.

apozharski: I personally actually don't like keeping multiple ways to do this as it is unclear for the user (and me :sweat_smile: ) as to which should have priority. In practice my preference is to have the plugin be `'relaxation'` and the relaxation type passed through `solver_options`.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Something to discuss: Is it better to throw errors or to quietly (I consider a warning quite as on observation outside a production environment 99% of people will simply ignore any and all warnings :sweat_smile:) change the option to a valid one and proceed. I have seen published papers in the past where the settings written in the paper make no sense and the tool that the authors used modified them with a warning and this makes for a lot of headaches. 

As such it may be clear that I prefer to "loudly" error out and tell the user what they should do. At the same time I know that there are people who simply shut down when confronted with red text and that can scare potential users away.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Technically the nice part of the vdx backend is changing the numbers isn't that hard after the fact (if a little detailed)
```
ocp_solver.discrete_time_problem.w.not_c_primals(1,2).init  = 5;
```
for example. 

However in this case there do need to be changes here made to allow for `n_c  > 1`.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Agreed. Added a TODO, that perhaps we also want to do some smarter re-ordering for a better structure.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: `f_ijk` and `q_ijk` is better, agreed.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: I have not implemented lifting as of now as I have not seen any major difficulties from even nonlinear $c(x)$. However, it is probably a good idea to provide as an option.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: I actually don't love having the initial values in the options, and I think having a robust set method is better than per-defining these things. As for not initializing at (0,0) I have actually observed the opposite for some of the PDS problems which is why the current initialization is what it is.

examples/new_design_examples/new_design_pds.m
nurkanovic: I also dont like it, but from time to time I am ok for exceptions. :D
but if we both agree that setting the plugi in via `solver_options` like all other options, then this is the way to go :D :D :D

+nosnoc/+model/Pds.m
nurkanovic: Hence, it is also undocumented. I prefer to keep it out of `main`, except it is needed for our current examples. 
But a few lines of comments explaining what it does (to help me and others) would be very welcome!

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: For now, I am agnostic about where we put the default lies aslong it is consistent.  
I just dont like them being hard-coded as our wisdom regarding this stuff is quite dynamic. Of course, whatever value leads to good perfomance is fine.
They are not hard coded in the Stewart and Heavside, so I suggest making it consistent.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: No difficulties, but it may improve performance. I am curious to learn about the experiences. 
Our guiding principle in general is anyway: affine functions into complementarity constraints.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: I think a good resolving compromise for this is: Error + instructive message why and what to use.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: But we dont expect everyone interested in only in the fronted to know the internally used vdx syntax. 
As for any bounds, initial values and co, I prefer having variables stored somewhere (This is also consistent with the currently new design existing in main)

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Yep there was a bug causing wrong cross complementarities for this mode. Fixed now.

examples/new_design_examples/new_design_msv.m
apozharski: I think we should return the value of `discrete_time_problem.p.T` (as the user may have modified it after the fact and it may not equal `model.T` but otherwise I agree and this is implemented in 'nosnoc.ocp.Solver.get'.

examples/new_design_examples/new_design_msv.m
apozharski: :+1:  I am happy to tear out the current implementation as I agree it is not ready for `main.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: This particular issue will be gone due to above rbp discussion. However I generally agree that backend vdx syntax should be hidden. As such I have a TODO to make a robust `.set` method for `nosnoc.Integrator` and `nosnoc.ocp.Solver` which should hide the vdx syntax.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Removed, per rbp discussion.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: In prior implementation this was explicitly disallowed. I spent some time looking into it and I am under the impression that is is not possible due to their structure, but I am not 100% on this.

examples/new_design_examples/new_design_msv.m
apozharski: I believe this is fixed. At least on my end it seems to work correctly.

examples/new_design_examples/new_design_pds.m
apozharski: I think setting the plugin the way it is now is mostly to have `mpccsol` have a similar syntax to `nlpsol`. I created an issue to box up this discussion: #109.

+nosnoc/+model/Pds.m
apozharski: We could alternatively drop the example that needs it (the sweeping set example). I have no strong opinions either way at the moment.

nurkanovic: Just add a small  comment in the src for what is `E` used and when it is needed.
But I would make `E` an optional thing you pass. E.g., if it is a "regular" PDS, then E is default identity, and you dont even need to pass it.

examples/new_design_examples/new_design_pds.m
apozharski: This is due to  me leaving GL as the `rk_scheme` which is now currently unsupported per rbp discussion.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Yep  this is now implemented. I am happy to make it a default as it does actually seem to improve convergence a bit :)

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: I assume we have the same in Stewart and Heaviside 🥹

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Yea I agree it may make sense to re-think exactly how the step equilibration is done. There are several ways to slice the existing modes. However I think this is probably a separate PR. As an aside the other mode now work for GCS too, them not working was due to a rebasing bug.

+nosnoc/+dcs/Gcs.m
apozharski: Yea this is not actually used in the discrete time problem, and was left purely due to copy paste. Removing now.

+nosnoc/+dcs/Gcs.m
apozharski: Partially yes. In this case this is because in a GCS we don't have any non-complementarity components that are used to define the algebraics $\lambda$, hence the algorithmic algebraic function is empty.

+nosnoc/+dcs/Gcs.m
apozharski: Yes precisely this. We can do this via depends on but it is some effort to add this improvement hence the TODO.

+nosnoc/+dcs/Gcs.m
apozharski: The property group printer uses the fields of the property group to populate what is displayed when default print is called for the object (for example if you type `dcs` in the interactive prompt). Adding the rest of the variables as for stewart now.

+nosnoc/+dcs/Gcs.m
apozharski: No reason. This should not be there.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Yea there is very little overlap. In theory we could move some of  the extremely common code into a base class but the 1% reduction in repetition is probably not worth the readability penalty. I am open to changing this in the future but for now I am happy as it is.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Probably should be an issue. Added #110

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Also fixed :smile:

+nosnoc/+dcs/Gcs.m
nurkanovic: Then it makes sense to not have it at all, instead of keeping track of any empty think.
Except, algorithmic algebraic conditions appear in the foreseeable future.

+nosnoc/+dcs/Gcs.m
nurkanovic: I am not sure should  we make all these TODOs Issues. I am afraid that we forget them if they live in stable code that we don't look at often.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: I agree.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: Lifted default is our friend (although one might check offline is it affine, this is the only case where lifting is an overkill).

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: Hence, lets make another issue :D 
At least the new issues are all quite small things (which will make review easy :D)

apozharski: Added issue: #111

+nosnoc/+dcs/Gcs.m
apozharski: Added issue #112

examples/new_design_examples/plot_moving_set.m
apozharski: PDS examples are now in the main examples folder :smile:

nurkanovic (approved):
All comments have been addressed. GJ!

PR #101, opened 2024-05-23
New design CLS Pipeline

### LINKED ISSUE #94: Implement new pipeline design for FESD-J reformulation of CLS.

apozharski:
resolves #94


examples/new_design_examples/new_design_cls2.m
nurkanovic, line 1: The integrator complains in the first steps that it has failed, despite the last nlpsol call was successful and the comp tolerance being very low (below 1e-14). Something is out of sync with tolerances and termination criteria?

nurkanovic:
The option `problem_options.lift_velocity_state = 1` leads to a failure. (Btw its purpose was to avoid symbolic inversion of the inertia matrix? By having   dot{v}  = z $ and  M z = f_x + contact_forces? We should add this simple explanation into the documentation in case I recall its purpose correctly.

+nosnoc/+ocp/Solver.m
```diff
@@ -44,7 +44,11 @@
                 obj.discrete_time_problem = nosnoc.discrete_time_problem.Heaviside(obj.dcs, opts);
                 obj.discrete_time_problem.populate_problem();
               case "nosnoc.model.Cls"
-                error("not implemented")
```
nurkanovic, line 47: nice change 🥇

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+            end
+
+            % 1d vars: Variables that are defined per control stage
+            obj.w.u(1:opts.N_stages) = {{'u', dims.n_u}, model.lbu, model.ubu, model.u0};
+            obj.p.p_time_var(1:opts.N_stages) = {{'p_time_var', dims.n_p_time_var}, model.p_time_var_val};
+            if opts.use_speed_of_time_variables && opts.local_speed_of_time_variable
+                obj.w.sot(1:opts.N_stages) = {{'sot', 1}, opts.s_sot_min, opts.s_sot_max, opts.s_sot0};
+            end
+
+            % TODO(@anton) This _severely_ hurts performance over the vectorized assignment by doing N_stages vertcats of
```
nurkanovic, line 54: we have this comment in all discrete_time_problem classes, but this is more of a vdx issue?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                    lbh = (1 - opts.gamma_h) * h0;
+                    if opts.time_rescaling && ~opts.use_speed_of_time_variables
+                        % if only time_rescaling is true, speed of time and step size all lumped together, e.g., \hat{h}_{k,i} = s_n * h_{k,i}, hence the bounds need to be extended.
+                        ubh = (1+opts.gamma_h)*h0*opts.s_sot_max;
+                        lbh = (1-opts.gamma_h)*h0/opts.s_sot_min;
+                    end
+                    obj.w.h(ii,1:opts.N_finite_elements(ii)) = {{'h', 1}, lbh, ubh, h0};
+                end
+                if obj.opts.step_equilibration == StepEquilibrationMode.linear_complementarity
+                    % TODO fix this :)
```
nurkanovic, line 74: ?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                end
+                % 2d Impulse vars
+                obj.w.Lambda_normal(ii,start_fe:opts.N_finite_elements(ii)) = {{'Lambda_normal', dims.n_c}, 0, inf, 1};
+                obj.w.P_vn(ii,start_fe:opts.N_finite_elements(ii)) = {{'P_vn', dims.n_c}, 0, inf, 1};
+                obj.w.N_vn(ii,start_fe:opts.N_finite_elements(ii)) = {{'N_vn', dims.n_c}, 0, inf, 1};
+                obj.w.Y_gap(ii,start_fe:opts.N_finite_elements(ii)) = {{'Y_gap', dims.n_c}, 0, inf, 1};
+                if model.friction_exists
+                    switch opts.friction_model
+                      case 'Polyhedral'
+                        obj.w.Lambda_tangent(ii,start_fe:opts.N_finite_elements(ii)) = {{'Lambda_tangent', dims.n_tangents}, 0, inf, 1};
```
nurkanovic, line 97: I am not a big fan of hard coded default values and bounds (on the other hand there are lot of these variables...
In fact, finite upper bounds here might help.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                    opts.rk_representation == RKRepresentation.differential_lift_x)
+                    obj.w.v(ii,1:opts.N_finite_elements(ii),1:opts.n_s) = {{'v', dims.n_x}};
+                end
+
+                if opts.lift_velocity_state
+                    obj.w.z_v(ii,1:opts.N_finite_elements(ii),1:opts.n_s) = {{'z_v', dims.n_x}};
+                end
+                
+                obj.w.z(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
+                obj.w.lambda_normal(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'lambda_normal', dims.n_c}, 0, 50, 0};
```
nurkanovic, line 161: quite arbitrary upper bound on lambda normal (and a quite low value). 
Take a particle with a mass of 50kg and you have an infeasible problem.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                obj.w.y_gap(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'y_gap', dims.n_c}, 0, inf, 0};
+                
+                if model.friction_exists
+                    switch opts.friction_model
+                      case 'Polyhedral'
+                        obj.w.lambda_tangent(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'lambda_tangent', dims.n_tangents}, 0, 50, 0};
```
nurkanovic, line 167: upper bound not justified.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                        obj.w.gamma_d(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'gamma_d', dims.n_c}, 0, inf, 0};
+                        obj.w.beta_d(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'beta_d', dims.n_c}, 0, inf, 1};
+                        obj.w.delta_d(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'delta_d', dims.n_tangents}, 0, inf, 1};
+                      case 'Conic'
+                        obj.w.lambda_tangent(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'lambda_tangent', dims.n_tangents}, -20, 20, 0};
```
nurkanovic, line 172: the bounds are to small

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                end
+
+                if ~opts.x_box_at_fe
+                    obj.w.x(1:opts.N_stages,1:(opts.N_finite_elements(ii)-1),opts.n_s+rbp).lb = -inf*ones(dims.n_x, 1);
+                    obj.w.x(1:opts.N_stages,1:(opts.N_finite_elements(ii)-1),opts.n_s+rbp).ub = inf*ones(dims.n_x, 1);
+                end
+            end
+
+            % build z_alg, z_impulse, and z_alg_f_x variable groups
+            z_alg = {obj.w.lambda_normal, obj.w.y_gap};
```
nurkanovic, line 202: explain the difference between the variable groups with a short comment.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+            p_global = obj.p.p_global();
+
+            x_0 = obj.w.x(0,0,opts.n_s);
+            z_0 = obj.w.z(0,0,opts.n_s);
+            
+            obj.g.z(0,0,opts.n_s) = {dcs.g_z_fun(x_0, z_0, obj.w.u(1), v_global, [p_global;obj.p.p_time_var(1)])};
+            
+            x_prev = obj.w.x(0,0,opts.n_s);
+            for ii=1:opts.N_stages
+                h0 = obj.p.T().val/(opts.N_stages*opts.N_finite_elements(ii));
```
nurkanovic, line 255: h_0

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                
+                ui = obj.w.u(ii);
```
nurkanovic, line 257: u_i (for consistency, we had a similar situation in the GCS/PDS PR)

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+
+                sum_h = 0;
+                for jj=1:opts.N_finite_elements(ii)
+                    if obj.opts.use_fesd
+                        h = obj.w.h(ii,jj);
+                        sum_h = sum_h + h;
+                    elseif opts.time_optimal_problem && ~opts.use_speed_of_time_variables
+                        h = obj.w.T_final()/(opts.N_stages*opts.N_finite_elements(ii));
+                    else
+                        h = h0;
```
nurkanovic, line 283: if this is the first time where use_fesd is false, then invoke an error message and ask the user to set use_fesd = true, as discussed in a previous comment the false case does not make sense.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+            % TODO(@anton) This _severely_ hurts performance over the vectorized assignment by doing N_stages vertcats of
+            %              casadi symbolics vs just a vectorized assignment which does one. As such there needs to be backend
+            %              work done for vdx to cache vertcats of SX somehow. Current theory is one can simply keep a queue of
+            %              symbolics to be added in a cell array until a read is done, at which point we call a single vertcat
+            %              on the whole queue which is _significantly_ faster.
+            % 2d vars: Variables that are defined for each finite element.
+            for ii=1:opts.N_stages
+                % other derived values
+                h0 = opts.h_k(ii);
+                if obj.opts.use_fesd
```
nurkanovic, line 63: if this is the first time where use_fesd is false, then invoke an error message and ask the user to set use_fesd = true, as discussed in a previous comment the false case does not make sense.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+            obj.g.z(0,0,opts.n_s) = {dcs.g_z_fun(x_0, z_0, obj.w.u(1), v_global, [p_global;obj.p.p_time_var(1)])};
+            
+            x_prev = obj.w.x(0,0,opts.n_s);
+            for ii=1:opts.N_stages
+                h0 = obj.p.T().val/(opts.N_stages*opts.N_finite_elements(ii));
+                
+                ui = obj.w.u(ii);
+                p_stage = obj.p.p_time_var(ii);
+                p = [p_global;p_stage];
+                if obj.opts.use_speed_of_time_variables && opts.local_speed_of_time_variable
```
nurkanovic, line 260: having `problem_options.time_optimal_problem = 1` works (can be tested on `new_design_cls_discs.m`

having `use_speed_of_time_variables = 1` (and `local_speed_of_time_variable = 1`)  leads to some noninstructive errors.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                        obj.g.f_c_eps(ii,jj) = {dcs.f_c_fun(x_eps), 0, inf};
+                    end
+                    
+                    switch opts.rk_representation
+                      case RKRepresentation.integral
+                        % In integral representation stage variables are states.
+                        x_ij_end = opts.D_rk(1)*x_lbp;
+                        for kk=1:opts.n_s
+                            x_ijk = obj.w.x(ii,jj,kk);
+                            z_ijk = obj.w.z(ii,jj,kk);
```
nurkanovic, line 312: just to double check, these are user algebarics (and we have not CLS examples where they appear?)?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                            if opts.g_path_at_stg
+                                obj.g.path(ii,jj,kk) = {dcs.g_path_fun(x_ijk, z_ijk, ui, v_global, p), model.lbg_path, model.ubg_path};
+                            end
+                            if size(model.G_path, 1) > 0
+                                G_path = dcs.G_path_fun(x_ijk, z_ijk, ui, v_global, p);
+                                H_path = dcs.H_path_fun(x_ijk, z_ijk, ui, v_global, p);
+                                obj.G.path(ii,jj,kk) = {G_path};
+                                obj.H.path(ii,jj,kk) = {H_path};
+                            end
+                            if opts.cost_integration
```
nurkanovic, line 338: what happens if opts.cost_integration =  false? 
we have no objective?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                            
+                            if opts.g_path_at_stg
+                                obj.g.path(ii,jj,kk) = {dcs.g_path_fun(x_ijk, z_ijk, ui, v_global, p), model.lbg_path, model.ubg_path};
+                            end
+                            if size(model.g_comp_path, 1) > 0
+                                g_comp_path = dcs.g_path_fun(x_ijk, z_ijk, ui, v_global, p);
+                                obj.G.path(ii,jj,kk) = {g_comp_path(:,1)};
+                                obj.H.path(ii,jj,kk) = {g_comp_path(:,2)};
+                            end
+                            if opts.cost_integration
```
nurkanovic, line 389: what happens if cost_integration false?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                    obj.f = obj.f + t_stage*dcs.f_lsq_u_fun(obj.w.u(ii),...
+                        model.u_ref_val(:,ii),...
+                        p);
+                end
+                if ~opts.cost_integration
+                    obj.f = obj.f + dcs.f_q_fun(x_ijk, z_ijk, ui, v_global, p);
+                end
+
+                % Clock <Constraints
+                % TODO(@anton) HERE BE DRAGONS. This is by far the worst part of current nosnoc as it requires the discrete problem
```
nurkanovic, line 489: I am not sure i unterstand this comment?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                if obj.opts.use_fesd && opts.equidistant_control_grid
+                    if opts.time_optimal_problem
+                        if opts.use_speed_of_time_variables
+                            obj.g.equidistant_control_grid(ii) = {[sum_h - opts.h;sot*sum_h - obj.w.T_final()/opts.N_stages]};
+                        else
+                            obj.g.equidistant_control_grid(ii) = {sum_h - obj.w.T_final()/opts.N_stages};
+                        end
+                    else
+                        if opts.relax_terminal_numerical_time
+                            % TODO(@armin) why is this allowed to be negative?
```
nurkanovic, line 500: may help convergence as the problem is "more relaxed"

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                obj.w.s_terminal_ell_inf = {{'s_terminal_ell_inf', 1}, 0, inf, 1e3};
+
+                g_terminal = [g_terminal-model.lbg_terminal-obj.w.s_terminal_ell_inf();
+                    -(g_terminal-model.ubg_terminal)-obj.w.s_terminal_ell_inf()];
+                obj.g.terminal = {g_terminal, -inf, 0};
+                obj.f = obj.f + obj.p.rho_terminal_p()*obj.w.s_terminal_ell_inf();
+            end
+        end
+
+        % TODO(@anton) is there a way to make this less horribly ugly?
```
nurkanovic, line 557: Maybe we should not judge too much in the public code. :P

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+
+            % Impulse complementarities
+            if opts.no_initial_impacts
+                start_fe = 2;
+            else
+                start_fe = 1;
+            end
+            for ii=1:opts.N_stages
+                for jj=start_fe:opts.N_finite_elements(ii)
+                    Gij = {};
```
nurkanovic, line 574: `G_ij` and so on, for sake of consistency.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                                Hij = [Hij,{obj.w.N_vt(ii,jj)}];
+                            end
+                        end
+                    end
+                    obj.G.impulse_comp(ii,jj) = {vertcat(Gij{:})};
+                    obj.H.impulse_comp(ii,jj) = {vertcat(Hij{:})};
+                end
+            end
+            % contact complementarities
+            if opts.use_fesd
```
nurkanovic, line 609: we may not even need this if, as false is meaningless and caught earlier.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                                            Hij = vertcat(Hij, {sum_n_vt});
+                                        end
+                                    end
+                                end
+                            end
+                            obj.G.cross_comp(ii,jj) = {vertcat(Gij{:})};
+                            obj.H.cross_comp(ii,jj) = {vertcat(Hij{:})};
+                        end
+                    end
+                  case CrossCompMode.FE_FE
```
nurkanovic, line 922: the good news is. all cross comp modes work, the disc manipulation example converges super fast with any of the options. :)

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                                        Hij = vertcat(Hij, {sum_n_vt});
+                                    end
+                                end
+                            end
+                            obj.G.cross_comp(ii,jj) = {vertcat(Gij{:})};
+                            obj.H.cross_comp(ii,jj) = {vertcat(Hij{:})};
+                        end
+                    end
+                end
+            else
```
nurkanovic, line 1016: this below is without fesd, we can drop it?

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                        end
+                        eta_vec = [eta_vec;eta];
+                        obj.eta_vec = eta_vec;
+                        delta_h = obj.w.h(ii,jj) - obj.w.h(ii,jj-1);
+                        homotopy_eq = [eta*delta_h - sigma;eta*delta_h + sigma];
+                        obj.g.step_equilibration(ii,jj) = {homotopy_eq, [-inf;0], [0;inf]};
+                    end
+                end
+                %obj.eta_fun = Function('eta_fun', {obj.w.sym}, {eta_vec});
+              case StepEquilibrationMode.linear_complementarity % TODO(@anton) implement this though we already have such pain w.r.t solving it may not be super useful
```
nurkanovic, line 1432: having this option acitve the code does not get this far. it fails before.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+            import casadi.*
+            model = obj.model;
+            opts = obj.opts;
+            dims = obj.dcs.dims;
+            
+            if ~opts.use_fesd % do nothing
+                return
+            end
+            rbp = ~opts.right_boundary_point_explicit;
+            % TODO (@anton) maybe unify some of these calculations
```
nurkanovic, line 1087: this would be nice, as there is much repetition. and we may drop support for direct (and even direct homotopy?) as they are not the choice to be used in practice.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: The details yea are perhaps a VDX issue but it would be good to keep a note here for reference regarding performance. I plan on fixing the performance issue in VDX next week and we can remove this.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Correct. Generally user algebraics in CLS are probably a very easy way to break ones model but gives us some flexibility.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Currently `linear_complementarity`is not implemented in `Cls` as it is not entirely clear to me how it should be implemented. I think this should be a todo. I moved the more useful todo here too.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: I would generally agree. Though like you said there are _a lot_ of defaults. I am ok with adding these to `Options`.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Correct. This is a vestige of my testing. will replace with max in options.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -0,0 +1,1600 @@
...
+                    z_impulse_ij = obj.z_impulse(ii,jj);
+
+                    if (jj ~= 1 || ~opts.no_initial_impacts)
+                        obj.g.impulse(ii,jj) = {dcs.g_impulse_fun(q_lbp,v_lbp,v_prev,z_impulse_ij, v_global, p)};
+                    else
+                        obj.g.v_continuity(ii,jj) = {v_lbp-v_prev};
+                    end
+                    % additional y_eps constraint
+                    if opts.eps_cls > 0
+                        x_eps = vertcat(q_lbp + h * opts.eps_cls * v_lbp, v_lbp);
```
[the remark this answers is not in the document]
apozharski: I believe some strange results occur if h is allowed to be smaller than the fixed `eps_cls` however and in experimenting the benefit seems nonexistant. I can make this an option however.

+nosnoc/+discrete_time_problem/Cls.m
apozharski:
```matlab
if ~opts.cost_integration
    obj.f = obj.f + dcs.f_q_fun(x_ijk, z_ijk, u_i, v_global, p);
end
```

`cost_integration = false` yields what can be seen as a pointwise cost at the end of each control stage. This can be useful for e.g. MLE of parameters etc. 

We can rename but I think the last time we had this conversation we didn't have a better name.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Basically this is a leak in the interface between the `model` and the `discrete_time_problem` here. Timefreezing is strictly a model concern however with the current naming scheme it looks like the discrete time problem needs to know whether a time freezing reformulation happened which is not the cleanest thing. In practice this is maybe dealt with renamed flag in `Options` called something like `has_physical_clock_state`, which is more generic.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Eh, I will drop this but considering the similar comments that exist in other projects (IPOPT, etc  just in our field) I wouldn't worry too much about it.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Answered above.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: If we are dropping non-fesd support for Cls we can probably catch that at the `ocp.Solver` stage.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Yep, moved the error to `ocp.Solver` and got rid of `use_fesd` checks in `discrete_time_problem.Cls`

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Ahhh  I see, It is not necessary in CLS. removed.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Dealt with via new options.

+nosnoc/+discrete_time_problem/Cls.m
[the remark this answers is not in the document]
apozharski: Option added.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: I think this should be done in a pass that resolves #111 but yes, agreed.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: This error is fixed. Was a typo.

+nosnoc/+discrete_time_problem/Cls.m
nurkanovic: this is equivalent to a explicit Euler cost integration with a constant step size one. And we can name the option 
`Euler_cost_integration` ( hence `opts.Euler_cost_integration= true` is same as `opts.cost_integration = false`)` and explain in options what it does with a comment when to use it (MLE)

examples/new_design_examples/new_design_cls2.m
apozharski: This is actually "expected" behavior. This example gives `N_homotopy=7` ,`sigma_0`,  and `complementarity_tol = 1e-6` which cannot be achieved.

+nosnoc/+discrete_time_problem/Cls.m
apozharski: Rename done and added more to documentation in `Options.m`

nurkanovic (approved):
LGTM 🤖

Issue #102, opened 2024-05-28
Clock state for integration grid and block band structure in FESD


nurkanovic:
Currently, we have for every FE an h. 
Alternatively, we can define a clock state t, and have a "control variable" h_n to determine the length of a finite element.
We simply integrate the clock state as t_{n+1} = t_n  + h_n;  and enforce a "terminal" constraint 
t_N  = T_final instead of sum_n  h_n  = T_final
The intermediate variables t_n help eliminate sum_n  h_n  = T_final, which kills the block banded structure.

PR #106, opened 2024-06-19
New design time freezing

### LINKED ISSUE #71: Stewart reformulation for time-freezing
Currently time-freezing only supports the Heaviside-step reformulation. Add functionality to be able to use the Stewart reformulation as well.

### LINKED ISSUE #85: Add option `relax_terminal_physical_time` for time_freezing problems
We have this for the plain numerical time, to make an ell_1 relaxation of the terminal time constraint. 
It would be nice to have the same also for physical time constraints (maybe even also for the stagewise constraints - but to be discussed).

I want also to test with this the ease of use of vdx, once we migrated to it.

apozharski:
Resolves #71. Resolves #85.


nurkanovic:
In simulation problems, we have always one stage, and 
`problem_options.impose_terminal_phyisical_time = 1;`
`problem_options.stagewise_clock_constraint = 1;`
lead to the very same result, but we may have the same terminal constraint twice (also one relaxed, other as hard, depending on the relaxation settings, hence once we run the integrator, `problem_options.stagewise_clock_constraint` should always be false.

examples/new_design_examples/new_design_tf_simple.m
```diff
@@ -6,61 +6,69 @@
...
-problem_options.N_sim = N_sim;
-problem_options.n_s = 3;
-
 integrator = nosnoc.Integrator(model, problem_options, solver_options);
 [t_grid, x_res, t_grid_full, x_res_full] = integrator.simulate();
-% %
+
+%% plot algebraics
+figure
+if problem_options.dcs_mode == DcsMode.Stewart
```
nurkanovic, line 63: The lengths of theta/alpha (full or reduced) are totally our of sync with the leghts of the t_grids, hence it is hard to plot them.
Moreover, the theta/alpha seem to switch their value sometimes wheren they are not supposed to (even if the problem converged)... it is worth taking a look why.

examples/new_design_examples/new_design_tf_discs.m
```diff
@@ -25,46 +25,39 @@
...
 problem_options = nosnoc.Options();
 solver_options = nosnoc.solver.Options();
 model = nosnoc.model.Cls();
-% settings
+%% settings
 problem_options.rk_scheme = RKSchemes.RADAU_IIA;
-problem_options.n_s = 2;  % number of stages in IRK methods
-problem_options.time_freezing = 1;
-problem_options.cross_comp_mode = 7;
+problem_options.n_s = 2;  % number of stages in IRK methods (TODO@Anton, other variations, e.g. Stewart, with n_s>1 converge nicely, this setting very slow)
```
nurkanovic, line 44: as indicated in the comment, heaviside and costum lifted converges much slower than stewart or heaviside without costum lift for all other settings the same. initialization of lifting variables is the problem?

+nosnoc/+dcs/Heaviside.m
```diff
@@ -129,7 +132,7 @@ function generate_equations(obj, opts)
 
             obj.f_x_fun = Function('f_x', {model.x, model.z, obj.alpha, obj.lambda_n, obj.lambda_p, model.u, model.v_global, model.p}, {obj.f_x, model.f_q});
             obj.f_q_fun = Function('f_q', {model.x, model.z, obj.alpha, obj.lambda_n, obj.lambda_p, model.u, model.v_global, model.p}, {model.f_q});
-            obj.g_z_fun = Function('g_z', {model.x, model.z, model.u, model.v_global, model.p}, {model.g_z});
+            obj.g_z_fun = Function('g_z', {model.x, obj.alpha, model.z, model.u, model.v_global, model.p}, {model.g_z});
```
nurkanovic, line 135: here it is a bit hard to follow without comments. 
Once again, g_z are the user algebaric equations, and g_alg_ are the DCS algebarics (e.g. lp equations) - please clarify and comment a bit.
And why are all variables from the model, only the alpha from the object (comment also briefly in code)

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -35,6 +35,10 @@ function create_variables(obj)
             obj.w.v_global = {{'v_global',dims.n_v_global}, model.lbv_global, model.ubv_global, model.v0_global};
             if opts.use_speed_of_time_variables && ~opts.local_speed_of_time_variable
                 obj.w.sot = {{'sot', 1}, opts.s_sot_min, opts.s_sot_max, opts.s_sot0};
+                if opts.time_freezing
+                    obj.p.rho_sot = {{'rho_sot',1}, opts.rho_sot};
+                    obj.f = obj.f + obj.p.rho_sot()*(obj.w.sot()-1)^2;
```
nurkanovic, line 40: should the sot variable be different for every control stage? otherwise, the problem might be infeasible.

nurkanovic: in other words, we should not have local sot in tf... god this is complicated :D 
I suggest to throw an error here and inform the user to use `local_speed_of_time_variable = true` (which is default anyway)  if `time_freezing = true`.

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -58,6 +66,11 @@ function create_variables(obj)
             %              symbolics to be added in a cell array until a read is done, at which point we call a single vertcat
             %              on the whole queue which is _significantly_ faster.
             % 2d vars: Variables that are defined for each finite element.
+            if opts.use_fesd && opts.use_numerical_clock_state
```
nurkanovic, line 69: what is a numerical clock state? A clock state that is not frozen I guess, but what it is purpose? To have the block diagonal sparsity? Please comment here briefly, and add explanation here: https://github.com/nurkanovic/nosnoc/pull/106/files#diff-8b063ca3da251c42716694289c0a76214cd4575bc0fcb9b2ef58167320fa6c3aR233

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -91,6 +111,7 @@ function create_variables(obj)
             obj.w.z(0,0,opts.n_s) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
             obj.w.lambda_n(0,0,opts.n_s) = {{['lambda_n'], dims.n_lambda},0,inf};
             obj.w.lambda_p(0,0,opts.n_s) = {{['lambda_p'], dims.n_lambda},0,inf};
+            obj.w.alpha(0,0,opts.n_s) = {{['alpha'], dims.n_alpha},0, 1};
```
nurkanovic, line 114: I am wondering is this upper bound on alpha needed. It is also present in the complementarity constraints via 1-alpha, hence we have the constraint twice?

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -363,28 +399,49 @@ function generate_direct_transcription_constraints(obj)
...
+                end
+                
+                % Handle possible physical time
+                if opts.time_freezing && opts.stagewise_clock_constraint
+                    x0 = obj.w.x(0,0,opts.n_s);
+                    t0 = x0(end);
+                    x_stage_end = obj.w.x(ii, opts.N_finite_elements(ii), opts.n_s+rbp);
+                    t_stage_end = x_stage_end(end);
+
+                    relax = vdx.RelaxationStruct(opts.relax_terminal_physical_time.to_vdx, 's_physical_time', 'rho_physical_time');
```
nurkanovic, line 436: commenting the vdx syntax is helpful

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -363,28 +399,49 @@ function generate_direct_transcription_constraints(obj)
...
                 %              to understand something about the time-freezing reformulation which is ugly.
-                if obj.opts.use_fesd && opts.equidistant_control_grid
-                    if opts.time_optimal_problem
-                        if opts.use_speed_of_time_variables
-                            obj.g.equidistant_control_grid(ii) = {[sum_h - opts.h;s_sot*sum_h - obj.w.T_final()/opts.N_stages]};
-                        else
-                            obj.g.equidistant_control_grid(ii) = {sum_h - obj.w.T_final()/opts.N_stages};
-                        end
+                % handle numerical time
+                if opts.use_fesd && opts.equidistant_control_grid
```
nurkanovic, line 406: if we have time freezing, the equidistant control grid in numerical time has not really a meaning, but adds more constraints. this might slow down convergence? 
hence, does it make sense to always turn equidistant_control_grid off as soon as we have time freezing on?

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -363,28 +399,49 @@ function generate_direct_transcription_constraints(obj)
...
+
+                    if opts.use_numerical_clock_state
+                        curr_t = obj.w.numerical_time(ii,opts.N_finite_elements(ii));
+                        obj.g.equidistant_numerical_grid(ii) = {curr_t - ii*ecg_rhs, relax};
+                    else
+                        if ~opts.time_freezing
+                            obj.g.equidistant_numerical_grid(ii) = {s_sot*sum_h - ecg_rhs, relax};
                         else
-                            obj.g.equidistant_control_grid(ii) = {t_stage-sum_h};
+                            obj.g.equidistant_numerical_grid(ii) = {sum_h - ecg_rhs, relax};
```
nurkanovic, line 421: the syntax with relax is not super readable, please comment what is vdx doing

+nosnoc/+discrete_time_problem/Heaviside.m
nurkanovic, line 436: just to double check, the constraint relaxation syntax is already used  - no more explicit case disticntion neede?

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -399,36 +456,113 @@ function generate_direct_transcription_constraints(obj)
                     model.x_ref_end_val,...
                     p_global);
             end
+
+            %  -- Constraint for the terminal numerical and physical time (if no equidistant grids are required) --
```
nurkanovic, line 460: This part is quite complicated as there are many possible cases.
maybe it would be useful to write a single line comment after every if/else to summarize what is happening :D
otherwise one needs to think quite a bit

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -399,36 +456,113 @@ function generate_direct_transcription_constraints(obj)
...
-                                                  % TODO(@anton): this is as it was implemented before. should handle lb != ub?
-                obj.f = obj.f + obj.p.rho_terminal_p()*(g_terminal-model.lbg_terminal)'*(g_terminal-model.lbg_terminal);
-              case ConstraintRelaxationMode.ELL_INF % l_inf
-                obj.w.s_terminal_ell_inf = {{'s_terminal_ell_inf', 1}, 0, inf, 1e3};
-
-                g_terminal = [g_terminal-model.lbg_terminal-obj.w.s_terminal_ell_inf();
-                    -(g_terminal-model.ubg_terminal)-obj.w.s_terminal_ell_inf()];
-                obj.g.terminal = {g_terminal, -inf, 0};
-                obj.f = obj.f + obj.p.rho_terminal_p()*obj.w.s_terminal_ell_inf();
+            relax = vdx.RelaxationStruct(opts.relax_terminal_constraint.to_vdx, 's_terminal', 'rho_terminal');
```
nurkanovic, line 560: nice simplification, but as said above, a one-line comment should explain the constraint relaxation syntax

nurkanovic: possibly add a link to the vdx docu or example where it is introduced (if it exists)

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -94,7 +114,8 @@ function create_variables(obj)
             obj.w.x(0,0,opts.n_s) = {{['x_0'], dims.n_x}, model.x0, model.x0, model.x0};
             obj.w.z(0,0,opts.n_s) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
             obj.w.lambda(0,0,opts.n_s) = {{['lambda'], dims.n_lambda},0,inf, 1};
-            obj.w.mu(0,0,opts.n_s) = {{'mu', dims.n_mu},0,inf};
+            obj.w.theta(0,0,opts.n_s) = {{['theta'], dims.n_theta},0, 1, 1/dims.n_theta};
```
nurkanovic, line 117: is the upper bound on theta needed? will again lead to two identical columns in the Jacobian if 
a single theta_i = 1, 
as we have also \sum theta_i = 1

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -112,7 +133,7 @@ function create_variables(obj)
                 obj.w.z(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'z', dims.n_z}, model.lbz, model.ubz, model.z0};
                 obj.w.lambda(ii,1:opts.N_finite_elements(ii),1:(opts.n_s+rbp)) = {{'lambda', dims.n_lambda},0, inf, 1};
                 obj.w.theta(ii,1:opts.N_finite_elements(ii),1:(opts.n_s)) = {{'theta', dims.n_theta},0, 1, 1/dims.n_theta};
```
nurkanovic, line 135: reconsider need for upper bound on theta

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -376,29 +410,49 @@ function generate_direct_transcription_constraints(obj)
...
-                %              to understand something about the
-                %              time-freezing reformulation which is ugly. (It is also just a quite complicated thing)
-                if obj.opts.use_fesd && opts.equidistant_control_grid
-                    if opts.time_optimal_problem
-                        if opts.use_speed_of_time_variables
-                            obj.g.equidistant_control_grid(ii) = {[sum_h - opts.h;s_sot*sum_h - obj.w.T_final()/opts.N_stages]};
-                        else
-                            obj.g.equidistant_control_grid(ii) = {sum_h - obj.w.T_final()/opts.N_stages};
-                        end
+                %              to understand something about the time-freezing reformulation which is ugly.
```
nurkanovic, line 415: we should judge less the public code and have neutral tone comments either explaining stuff, or adding remarks (for TODOs we better have issues, as they are easier to find). 
also, the time options are just quite complicated.

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -376,29 +410,49 @@ function generate_direct_transcription_constraints(obj)
...
+                        obj.g.equidistant_numerical_grid(ii) = {curr_t - ii*ecg_rhs, relax};
+                    else
+                        if ~opts.time_freezing
+                            obj.g.equidistant_numerical_grid(ii) = {s_sot*sum_h - ecg_rhs, relax};
                         else
-                            obj.g.equidistant_control_grid(ii) = {t_stage-sum_h};
+                            obj.g.equidistant_numerical_grid(ii) = {sum_h - ecg_rhs, relax};
                         end
                     end
+                    if relax.is_relaxed
```
nurkanovic, line 435: explain syntax and consider using more instructive names.
relax.is_relaxed is maybe nonobvious

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -376,29 +410,49 @@ function generate_direct_transcription_constraints(obj)
...
+                end
+
+                % Handle possible physical time
+                if opts.time_freezing && opts.stagewise_clock_constraint
+                    x0 = obj.w.x(0,0,opts.n_s);
+                    t0 = x0(end);
+                    x_stage_end = obj.w.x(ii, opts.N_finite_elements(ii), opts.n_s+rbp);
+                    t_stage_end = x_stage_end(end);
+
+                    relax = vdx.RelaxationStruct(opts.relax_terminal_physical_time.to_vdx, 's_physical_time', 'rho_physical_time');
```
nurkanovic, line 447: as in heaviside, explain the vdx specific syntax here

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -413,38 +467,111 @@ function generate_direct_transcription_constraints(obj)
                     model.x_ref_end_val,...
                     p_global);
             end
+
+            %  -- Constraint for the terminal numerical and physical time (if no equidistant grids are required) --
+            % If the control grid is not equidistant, the constraint on sum of h happen only at the end.
+            % The constraints are split to those which are related to numerical and physical time, to make it easier to read.
+            if opts.time_freezing
```
nurkanovic, line 474: the code here is identical to the one in Heaviside.. not sure does it make sense to extract it into a function (or at least copy all the explanation comments (cf. comment in Heaviside on explaining every branching)

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -413,38 +467,111 @@ function generate_direct_transcription_constraints(obj)
...
+                    end
+                end
+            end
             
             % Terminal constraint
             if opts.relax_terminal_constraint_homotopy
-                % TODO: we should have an enum for treating terminal
-                % constraints (or relaxing constraints in general, cf. https://github.com/nurkanovic/nosnoc/issues/96
-                error("Currently unsupported.")
+                error("Currently unsupported")
```
nurkanovic, line 568: for sake of consistency `error("nosnoc: Currently unsupported")`

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -413,38 +467,111 @@ function generate_direct_transcription_constraints(obj)
...
-                     % TODO(@anton): this is as it was implemented before. should handle lb != ub?
-                obj.f = obj.f + obj.p.rho_terminal_p()*(g_terminal-model.lbg_terminal)'*(g_terminal-model.lbg_terminal);
-              case ConstraintRelaxationMode.ELL_INF % l_inf
-                obj.w.s_terminal_ell_inf = {{'s_terminal_ell_inf', 1}, 0, inf, 1e3};
-
-                g_terminal = [g_terminal-model.lbg_terminal-obj.w.s_terminal_ell_inf();
-                    -(g_terminal-model.ubg_terminal)-obj.w.s_terminal_ell_inf()];
-                obj.g.terminal = {g_terminal, -inf, 0};
-                obj.f = obj.f + obj.p.rho_terminal_p()*obj.w.s_terminal_ell_inf();
+            relax = vdx.RelaxationStruct(opts.relax_terminal_constraint.to_vdx, 's_terminal', 'rho_terminal');
```
nurkanovic, line 571: same remark as before, explain specifics

+nosnoc/+model/Base.m
```diff
@@ -342,20 +342,20 @@ function verify_and_backfill(obj, opts)
             if size(obj.g_path, 1) ~= 0
                 g_path_constraint  = 1;
                 dims.n_g_path = length(obj.g_path);
-                if size(obj.g_path_lb, 1) ~= 0
-                    if length(obj.g_path_lb)~=dims.n_g_path;
-                        error('The user provided vector g_path_lb has the wrong size.')
+                if size(obj.lbg_path, 1) ~= 0
+                    if length(obj.lbg_path)~=dims.n_g_path;
+                        error('The user provided vector lbg_path has the wrong size.')
```
nurkanovic, line 347: `error("nosnoc: ...`

nurkanovic: i corrected in the CLS branch i believe, so appropriate merging might it

+nosnoc/+model/Base.m
```diff
@@ -408,12 +408,12 @@ function verify_and_backfill(obj, opts)
 
                 % some custom handling for objective functions:
                 if strcmp(name, 'f_q')
-                    if obj.f_q == 0
+                    if class(obj.f_q) ~= "casadi.SX" && obj.f_q == 0
```
nurkanovic, line 411: ?

+nosnoc/+model/Cls.m
```diff
@@ -0,0 +1,180 @@
+classdef Cls < nosnoc.model.Base
```
nurkanovic, line 1: I made some edits in the CLS PR in this file (mostly docu, explanation).

Btw, stupid question, if we use time_freezing, do all CLS variables get created at all? All the Lambda, Y_gap and so on, or does it get removed (we dont need it for the later PSS).

+nosnoc/+ocp/Solver.m
```diff
@@ -44,7 +50,11 @@
                 obj.discrete_time_problem = nosnoc.discrete_time_problem.Heaviside(obj.dcs, opts);
                 obj.discrete_time_problem.populate_problem();
               case "nosnoc.model.Cls"
-                error("not implemented")
+                obj.dcs = nosnoc.dcs.Cls(model);
```
nurkanovic, line 53: 😎

+nosnoc/+ocp/Solver.m
```diff
@@ -98,37 +108,66 @@ function solve(obj, plugin)
             end
         end
 
+        function set_param(obj, param, value)
+        % TODO (@anton) figure out how to do a set with indexing
+            if ~obj.discrete_time_problem.p.has_var(param);
+                error(['nosnoc:' char(param) ' does not exist as a parameter for this OCP']);
```
nurkanovic, line 114: full stop at end of the sentence (also in line 60)

+nosnoc/Options.m
```diff
@@ -181,15 +184,15 @@
         %     This option is currently unimplemented.
         relax_terminal_constraint_homotopy(1,1) logical = 0; 
 
-        relax_terminal_numerical_time(1,1) logical = 0; %boolean: If true instead of imposing $\sum h = T$, add it as $\ell_1$ penalty term.
+        relax_terminal_numerical_time(1,1) ConstraintRelaxationMode = ConstraintRelaxationMode.NONE; %boolean: If true instead of imposing $\sum h = T$, add it as $\ell_1$ penalty term.
```
nurkanovic, line 187: comment outdated, not boolean anymore?

+nosnoc/Options.m
```diff
@@ -181,15 +184,15 @@
...
         rho_terminal_numerical_time(1,1) double {mustBeNonnegative} = 1e2 % double: Weight used to penalize terminal numerical time violation.
         
         % boolean: If True the terminal numerical time constraint violation penalty is governed by homotopy parameter
         %
         % Warning:
         %     This option is currently unimplemented
         relax_terminal_numerical_time_homotopy (1,1) logical = 0; % us the homotopy parameter for the penalty.
-        relax_terminal_physical_time(1,1) logical = 0; % instead of imposing $t(T) = T$, add it as $\ell_1$ penalty term.
+        relax_terminal_physical_time(1,1) ConstraintRelaxationMode = ConstraintRelaxationMode.NONE; % instead of imposing $t(T) = T$, add it as $\ell_1$ penalty term.
```
nurkanovic, line 195: docu comment outdated

+nosnoc/+time_freezing/cls_elastic.m
```diff
@@ -0,0 +1,74 @@
...
+    pss_model.dims = dims;
+
+    % transfer all common properties to pss_model
+    common_props = properties('nosnoc.model.Base');
+    for ii=1:length(common_props)
+        prop = common_props{ii};
+        pss_model.(prop) = cls_model.(prop);
+    end
+    
+    dims.n_contacts = length(cls_model.f_c);
```
nurkanovic, line 13: only n_contact=1 is supported for the automated elastic, right?

+nosnoc/+time_freezing/cls_elastic.m
```diff
@@ -0,0 +1,74 @@
...
+        else
+            pss_model.f_q_T = L;
+        end
+        dims.n_quad = 1;
+    end
+    
+    % uneven number of states = it is assumed that the clock state is defined.
+    t = define_casadi_symbolic(opts.casadi_symbolic_mode,'t',1);
+    % update lower and upper bounds of lbx and ubx
+    pss_model.lbx = [pss_model.lbx;-inf];
```
nurkanovic, line 43: we have natural lower bounds for the clock state, 
it must be nonnegative, but can put -1 for some relaxation
upper bound is also not much bigger than T (except in time optimal) but a suff large value, .e.g. 100 or 1000 might help convergence...

nurkanovic:
A small example with two contacts (either simulation or some disc manipulation ocp) would be helpful to check and verify the correctness of the multi contact implementation.

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+    common_props = properties('nosnoc.model.Base');
+    for ii=1:length(common_props)
+        prop = common_props{ii};
+        pss_model.(prop) = cls_model.(prop);
+    end
+    
+    dims.n_contacts = length(cls_model.f_c);
+    % check dimensions of contacts
+    if isempty(dims.n_dim_contact)
+        warning('nosnoc: Please n_dim_contact, dimension of tangent space at contact (1, 2 or 3)')
```
nurkanovic, line 17: can it be 3?

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+    % uneven number of states = it is assumed that the clock state is defined.
+    t = define_casadi_symbolic(opts.casadi_symbolic_mode,'t',1);
+    % update lower and upper bounds of lbx and ubx
+    pss_model.lbx = [pss_model.lbx;-inf];
+    pss_model.ubx = [pss_model.ubx;inf];
+    pss_model.x = [pss_model.x;t];
+    pss_model.x0 = [pss_model.x0;0];
+    
+    % normal and tangential velocities
+    eps_t = 1e-7;
```
nurkanovic, line 50: comment its significance, why isnt in options?

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+    v_normal = cls_model.J_normal'*cls_model.v;
+    if cls_model.friction_exists
+        if dims.n_dim_contact == 2
+            v_tangent = (cls_model.J_tangent'*cls_model.v)';
```
nurkanovic, line 54: why are the tangential velocities of 2 and 3 treated differently?
For example, there might be few discs in 2d plane, and there are multiple possible tangential velocities (and more might be active at time, e.g., one discs pushed a middle disc, the middle disch pushes the last one, to get it to its goal (a nice ocp example, where the contact between the very first (actuated) and last is not modeled)

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+        v_tangent  = [];
+    end
+    
+    % parameter for auxiliary dynamics
+    % TODO add to options
+    a_n = 100;
+    %% Time-freezing reformulation
+    % unconstrained dynamics with clock state
+    f_ode = [cls_model.v;...
+        cls_model.invM*cls_model.f_v;
```
nurkanovic, line 74: In CLS, we have the option to lift the velocity state for numerical inversion of M.
We may use this here as well (either make an issue, or fix it in this PR if it is not too much effort).

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+        f_quad;
+        1];
+
+    % Auxiliary dynamics
+    % where to use invM, in every auxiliary dynamics or only at the end
+    % TODO put this in problem opts
```
nurkanovic, line 80: it is mathematically the same, i guess the latter is easier to read and consistent with the papers.

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+        inv_M_aux = eye(dims.n_q);
+        inv_M_ext = blkdiag(zeros(dims.n_q),cls_model.invM,zeros(1+dims.n_quad));
+    else
+        inv_M_aux = cls_model.invM;
+        inv_M_ext = eye(dims.n_x+1);
+    end
+    f_aux_pos = []; % matrix with all auxiliary tangent dynamics
+    f_aux_neg = [];
+    % time freezing dynamics
+    if opts.stabilizing_q_dynamics
```
nurkanovic, line 91: it a sort of Baumgartner stabilization of first order term for (f_c(q)= 0). 
The idea is , if f_c(q) \neq 0 numerically in sliding mode, then there is vectorfield pushing it to it, or making sure it stays there. We are ready to describe this in the options documentation (it is not experimental anymore).

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+        f_q_dynamics = -opts.kappa_stabilizing_q_dynamics*cls_model.J_normal*diag(cls_model.f_c);
+    else
+        f_q_dynamics = zeros(dims.n_q,dims.n_contacts);
+    end
+    f_aux_normal = [f_q_dynamics;inv_M_aux*cls_model.J_normal*a_n;zeros(1+dims.n_quad, 1)];
+
+    if opts.nonsmooth_switching_fun
```
nurkanovic, line 98: rename option to time_freezing_nonsmooth_switching_fun and add describe in the options what it is (for docu). 
this is indeed an experimental thing (btw a way to simplify the description of the T-shape boundary of regions with less functions)

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+        pss_model.c = [max_smooth_fun(cls_model.f_c,v_normal,0);v_tangent];    
+    else
+        if dims.n_dim_contact == 2
+            pss_model.c = [cls_model.f_c;v_normal;v_tangent'];
+        else
+            pss_model.c = [cls_model.f_c;v_normal;v_tangent_norms-eps_t];
```
nurkanovic, line 104: Please add a reference to this choice and descirpition of eps_t in options (it is in the 3d friction section in the automatica paper i believe)

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+    for ii = 1:dims.n_contacts
+        if cls_model.friction_exists && cls_model.mu(ii)>0
+            % auxiliary tangent;
+            if dims.n_dim_contact == 2
+                v_tangent_ii = cls_model.J_tangent(:,ii)'*cls_model.v;
+                f_aux_pos_ii = [f_q_dynamics(:,ii) ;inv_M_aux*(cls_model.J_normal(:,ii)-cls_model.J_tangent(:,ii)*(cls_model.mu(ii)))*a_n;0]; % for v>0
+                f_aux_neg_ii = [f_q_dynamics(:,ii) ;inv_M_aux*(cls_model.J_normal(:,ii)+cls_model.J_tangent(:,ii)*(cls_model.mu(ii)))*a_n;0]; % for v<0
+            else
+                v_tangent_ii = v_tangent(:,ii);
+                f_aux_pos_ii = [f_q_dynamics(:,ii);inv_M_aux*(cls_model.J_normal(:,ii)*a_n-cls_model.J_tangent(:,ii*2-1:ii*2)*cls_model.mu(ii)*a_n*v_tangent_ii/norm(v_tangent_ii+1e-12));0]; % for v>0
```
nurkanovic, line 117: use also eps_t instead of 1e-12? or eps_t*1e-3?

+nosnoc/+time_freezing/cls_inelastic.m
```diff
@@ -0,0 +1,189 @@
...
+    end
+    if cls_model.friction_exists
+        f_aux = [f_aux_pos,f_aux_neg];
+        dims.n_aux = 2*dims.n_contacts;
+    else
+        f_aux = f_aux_normal;
+        dims.n_aux = dims.n_contacts;
+    end
+
+    % Build logical functions because its easier than hand picking logic.
```
nurkanovic, line 132: from here to the end, a few comments on the equations will improve readability

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: This is the clock state described in #102. Will add comment.

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: We don't include $1-\alpha$ in the generic constraints $g$. so the upper box constraint is the unique $1-\alpha \ge 0$ constraint

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: On the contrary including the numerical time constraints improves convergence in my experience. As discussed there isn't necessarily a good rigorous reason as to why this should be, but for now we keep it I think.

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: Yep. The automatic relaxation is done here in `vdx.ConstraintVector` https://github.com/apozharski/vdx/blob/e379f6336c9be1572aacafc8fbd239f7d4ccde13/%2Bvdx/ConstraintVector.m#L79-L163

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: This is an excellent idea, however I have not updated `vdx` docs recently. Will put it in vdx issues.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
+function heaviside_model = cls_inelastic_multicontact(cls_model, opts)
```
nurkanovic, line 1: add a brief description of the function that this is a tailored lifting procedure for Heaviside that exploits the specific structure of the pss. It might not be obvious for anyone outside.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
...
+    common_props = properties('nosnoc.model.Base');
+    for ii=1:length(common_props)
+        prop = common_props{ii};
+        heaviside_model.(prop) = cls_model.(prop);
+    end
+    
+    dims.n_contacts = length(cls_model.f_c);
+    % check dimensions of contacts
+    if isempty(dims.n_dim_contact)
+        warning('nosnoc: Please n_dim_contact, dimension of tangent space at contact (1, 2 or 3)')
```
nurkanovic, line 17: same comment apply here as in the cls_inelastic

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
...
+    % uneven number of states = it is assumed that the clock state is defined.
+    t = define_casadi_symbolic(opts.casadi_symbolic_mode,'t',1);
+    % update lower and upper bounds of lbx and ubx
+    heaviside_model.lbx = [heaviside_model.lbx;-inf];
+    heaviside_model.ubx = [heaviside_model.ubx;inf];
+    heaviside_model.x = [heaviside_model.x;t];
+    heaviside_model.x0 = [heaviside_model.x0;0];
+    
+    % normal and tangential velocities
+    eps_t = 1e-7;
```
nurkanovic, line 50: eps_t as option, same is in the other multicontact,

+nosnoc/+model/Cls.m
apozharski: No. the FESD-J vars get created in the `dcs` step of the pipeline. Time freezing occurs at the `model` stage taking `nosnoc.model.Cls` and replacing it with `nosnoc.model.Pss` or `nosnoc.model.Heaviside` and feeding that into the rest of the pipeline.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
...
+    end
+    if cls_model.friction_exists
+        f_aux = [f_aux_pos,f_aux_neg];
+        dims.n_aux = 2*dims.n_contacts;
+    else
+        f_aux = f_aux_normal;
+        dims.n_aux = dims.n_contacts;
+    end
+
+    % Build logical functions because its easier than hand picking logic.
```
nurkanovic, line 134: a lot of shared core. again the dilemma do we extract them into a function. i have no strong opinion.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
...
+                    alpha_aux(ii) = (1-alpha_q(ii))*(1-alpha_v_normal(ii));
+                end
+            end
+        end
+    end
+    heaviside_model.z = [theta;beta];
+    heaviside_model.g_z = [theta-[alpha_ode;alpha_aux];
+        beta - [beta_bilinear_ode_expr; beta_bilinear_aux_expr; beta_prod_expr]];
+    
+    f_all = [f_ode (inv_M_ext*f_aux)];
```
nurkanovic, line 288: here the inv_M_ext is used instead of inv_M on every auxiliary. I like the readiblity of the latter more, and it should not have any influence on the perfomance.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
```diff
@@ -0,0 +1,293 @@
...
+            alpha_v_tangent = alpha(2*dims.n_contacts+1:end);
+        end
+    else
+        alpha_qv = alpha(1:dims.n_contacts);
+        if cls_model.friction_exists
+            alpha_v_tangent = alpha(dims.n_contacts+1:end);
+        end
+    end
+    alpha_ode = 1;
+    alpha_aux = SX(zeros(dims.n_aux ,1));
```
nurkanovic, line 149: is SX vx MX here an issue?

nurkanovic (changes requested):
Looks fine except the few minor aestetics comments. 
As mentioned earlier, having a minimal two contacts example (like the OCP with two discs, described in some of the examples) would be helpful to verify correctness.

+nosnoc/+dcs/Heaviside.m
apozharski: Technically these are all documented (though probably need to be _better_ documented) in `nosnoc.dcs.Base`. I can add some brief documentation though I am always wary of documenting in multiple places as this has a tendency to run into desyncing issues.

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: You mean we should not have global sot in tf? I think we _need_ local sot in order to maintain feasibility in each control stage when `stagewise_clock_constraint`/`equidistant_control_grid` is true.

nurkanovic: indeed, i confused myself a bit. global is needed e.g. in simulation, as there is only one stage (but it is equivalent to local).
local is needed for `stagewise_clock_constraint =true`. so we on the other hand, global sot with multiple control stages might lead to infeasible problems. 

in summary i was questioning the need of global sot, as local does the job (and my be even necessary). think of two stage, one has a jump and the other doesnt, and stagewise clock constraint. it can be feasible only with local.

+nosnoc/+discrete_time_problem/Heaviside.m
nurkanovic: Does the option work reliably? E.g. in simple non-tf PSS problems?

+nosnoc/+discrete_time_problem/Heaviside.m
nurkanovic: If it improves convergence, then it has it justification. 
It restrictis additionaly the feasible set, might be just a non-obvious guiding constraint.

+nosnoc/+model/Cls.m
nurkanovic: perfect!

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: Yea it works, though the effect on convergence is mild.

examples/new_design_examples/new_design_tf_simple.m
apozharski: This is a hilarious bug. Each solution step was being added twice to the integrator history. whoops :upside_down_face:

+nosnoc/+time_freezing/cls_elastic.m
apozharski: Do we want to add options for these? I suspect the effects here would be mild, though the implementation is not too difficult so it may be worth it.
=

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
apozharski: Yea extracted into a few functions. The interfacing is not the cleanest but that is ok probably to trade for one source of truth.

+nosnoc/+time_freezing/cls_inelastic.m
apozharski: Added to comments.

+nosnoc/+time_freezing/cls_inelastic.m
apozharski: No. This is copy and pasted from the old tf code so :shrug:. removed.

+nosnoc/+time_freezing/cls_inelastic.m
apozharski: You mean 1 and 2? I don't quite understand the question.

+nosnoc/+time_freezing/cls_elastic.m
apozharski: Yep. added an error.

+nosnoc/+discrete_time_problem/Stewart.m
apozharski: Perhaps it may make sense to extract this though I am unsure. I think we can leave it as is for now and if having two copies is inconvenient we can extract then.

+nosnoc/+discrete_time_problem/Heaviside.m
apozharski: Added Comment

examples/new_design_examples/new_design_tf_discs.m
apozharski: I believe this is now fixed with changed initialization and adding initialization calculation for the lifting variables.

+nosnoc/+model/Base.m
apozharski: Comparing a double to and SX even if both are 0 does not work.  Though this is also subtly broken so I made the check different. This is so empty `f_q` `f_q_T` do not get displayed (as was requested when this feature was implemented).

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
apozharski: Unsure but it is not necessary to predefine this anyway. fixed.

+nosnoc/+time_freezing/cls_inelastic.m
apozharski: Added a few comments.

apozharski:
> Looks fine except the few minor aestetics comments. As mentioned earlier, having a minimal two contacts example (like the OCP with two discs, described in some of the examples) would be helpful to verify correctness.

I added the three cart problem as a test of the multicontact implementation. It seems to work.

+nosnoc/+time_freezing/cls_inelastic_multicontact.m
apozharski: Added a brief explanation, though you probably have a better understanding of the details here. I would need to reacquaint myself with exactly how the sum of $F_aux$ is equivalent.

+nosnoc/+discrete_time_problem/Stewart.m
apozharski: Explained in one comment in each Stewart and Heaviside.

+nosnoc/+discrete_time_problem/Stewart.m
apozharski: #117 replaces these comments.

+nosnoc/+time_freezing/cls_inelastic.m
apozharski: I think this is only possible easily in the multicontact formulation. I will add an issue to think about this some more.

apozharski:
> In simulation problems, we have always one stage, and `problem_options.impose_terminal_phyisical_time = 1;` `problem_options.stagewise_clock_constraint = 1;` lead to the very same result, but we may have the same terminal constraint twice (also one relaxed, other as hard, depending on the relaxation settings, hence once we run the integrator, `problem_options.stagewise_clock_constraint` should always be false.

Yep this is now checked in `Options.preprocess`

+nosnoc/+time_freezing/cls_inelastic.m
nurkanovic: In the end it is just lifting the r.h.s. whatever is in it.

nurkanovic (approved):

nurkanovic:
Thanks for addressing all my comments. We are almost ready to merge. 

I have a few final comments:

- simulation: Gauss Legendre leads to infeasible problems both with Stewart and Heaviside and both elastic and inelastic impacts, with GL size of algebraic e.g. theta and t_grid not consistent.
- on discs example Heaviside not-lifted much slower than lifted (are the initializations consistent - if yes, then just the nonlinearity makes the difference)
- disc example Heaviside with n_s = 1 breaks the plotting for some reason. 
- tf_simulation example and inelastic impacts, the plots of the the algebraic variables is quite strange (I tried Radau, as it converges)

apozharski:
> * simulation: Gauss Legendre leads to infeasible problems both with Stewart and Heaviside and both elastic and inelastic impacts, with GL size of algebraic e.g. theta and t_grid not consistent.

The bug that was causing infeasibility is fixed however size mismatch is expected as we do not include the rhs theta/alpha as a variable in the mpcc.

> * on discs example Heaviside not-lifted much slower than lifted (are the initializations consistent - if yes, then just the nonlinearity makes the difference)

Initialization for $\alpha$ is consistent between both.

> * disc example Heaviside with n_s = 1 breaks the plotting for some reason.

With heuristic step equilibration the solver finds a solution where each other h is zero. This breaks our plotting code.

> * tf_simulation example and inelastic impacts, the plots of the the algebraic variables is quite strange (I tried Radau, as it converges)

Does not seem strange to me. the theta multipliers and beta seem to be correct.

PR #124, opened 2024-08-28
[WIP] New design remaining examples

apozharski:
PR that brings remaining examples up to date to use the new design.

Also fixes and moves benchmarks to their new home.


nurkanovic:
LGTM, before merging
I have three small comments: 
 - there are some files in monoped_ocp (now time_freezing_monoped_ocp) that could also be moved to the benchmark folder?
 - msv.m please add a reference from where does this example come form
 - run_all_experiments in ivp_problem leads to a wrong output for fesd. is it a plotting/output error for the objective or some bug?

apozharski:
> LGTM, before merging I have three small comments:
> 
>     * there are some files in monoped_ocp (now time_freezing_monoped_ocp) that could also be moved to the benchmark folder?
> 
>     * msv.m please add a reference from where does this example come form
> 
>     * run_all_experiments in ivp_problem leads to a wrong output for fesd. is it a plotting/output error for the objective or some bug?

Hmmm yes the first two are fixed. The last one I need to think about because we do not separate the augmented objective and regular objective. I have been thinking of adding support for this kind of stuff to vdx (i.e. named sections of the objective function) perhaps that is a solution. The alternative is to use direct step eq mode.

nurkanovic (approved):
LGTM

PR #130, opened 2024-09-18
Implementing an MPC class which solves the full MPCC at each iteration.

apozharski:
Adds both a base class for other MPC algorithms and a basic implementation of one using `nosnoc.ocp.Solver` as the backend. 

Adds an example of using this MPC class for the turbo car model.


nurkanovic (approved):
Great work, love it!

PR #141, opened 2024-11-19
Finish moving to integrator namespace

apozharski:
Move FESD integrator to `nosnoc.integrator.FESD` but keep the frontend class `nosnoc.Integrator` which wraps the "plugins".  Also finalize the move to `nosnoc.integrator.Options` to house integrator options.


benchmarks/oscilator_integrator_order/integrator_order_experiment.m
```diff
@@ -64,7 +65,7 @@
     solver_options.complementarity_tol = complementarity_tol;
     solver_options.N_homotopy = 40; % number of steps
     solver_options.homotopy_update_slope = 0.05; % decrease rate
-    solver_options.use_previous_solution_as_initial_guess = 1; % warm start integrator
+    % solver_options.use_previous_solution_as_initial_guess = 1; % warm start integrator
```
nurkanovic, line 68: this lives now in integrator options and causes a bug. this bug is present in most examples where this option is changed

benchmarks/oscilator_integrator_order/integrator_order_experiment.m
```diff
@@ -100,7 +101,7 @@
             problem_options.N_sim = N_sim;
             % generate new model with updated settings;
             model = oscilator(struct);
-            integrator = nosnoc.Integrator(model, problem_options, solver_options);
+            integrator = nosnoc.Integrator(model, problem_options, integrator_options);
             [t_grid, x_res, t_grid_full, x_res_full] = integrator.simulate();
             stats = integrator.stats;
```
nurkanovic, line 106: broken, as the method stats does not exist in integrator (and is rather part of solver which was not transfered accordingly)

examples/oscilator_simulation/oscilator_simulation_example.m
nurkanovic, line 1: Sever redundancy. If something exists in the benchmark folder, then it should be removed from the examples.

examples/oscilator_simulation/smoothed_pss.m
nurkanovic, line 1: 1) It is redundant to have two files for smoothed integrator (this one) and for the FESD integrator, in `osiclator_simulation_example`  - it should be one file where just the integrator options are changed.
2) Somethings if wrong with creating the Heaviside dcs when one calls the smoothed integrator, in particular, the dims are empty, causing the following error message
```
Dot indexing is not supported for variables of this type.

Error in nosnoc.dcs.Heaviside/generate_variables (line 49)
                dims.n_alpha = sum(obj.dims.n_c_sys); % number of modes

Error in nosnoc.integrator.SmoothedPss (line 37)
                obj.dcs.generate_variables(opts);

Error in smoothed_pss (line 87)
integrator = nosnoc.integrator.SmoothedPss(model, problem_options, integrator_options);
```

examples/oscilator_simulation/oscilator_simulation_example.m
```diff
@@ -88,7 +90,7 @@
 F = [f_11 f_12];
 model.F = F;
 %% Call integrator
-integrator = nosnoc.Integrator(model, problem_options, solver_options);
+integrator = nosnoc.Integrator(model, problem_options, integrator_options);
```
nurkanovic, line 91: this is the default call for an FESD integrator, but if I want a smoothed I need:
`integrator = nosnoc.integrator.SmoothedPss(model, problem_options, integrator_options);`

I would strongly prefer a unified syntax, i.e., an FESD integrator should be created via 
`integrator = nosnoc.integrator.FESD(model, problem_options, integrator_options);`

and so on.. otherwise it gets quickly too confusing...

+nosnoc/Integrator.m
nurkanovic, line 1: does it make sense to remove this and completely rely on the new integrator implementation with its diverse plugins (FESD, smoothed, and others that will follow)?

examples/oscilator_simulation/oscilator_simulation_example.m
nurkanovic: or maybe the better option is to have:
`integrator = nosnoc.Integrator(model, problem_options, integrator_options);`
and select the plugin via `integrator_options` . . .

nurkanovic: apparently, both are avilable and `nosnoc.Integrator` works fine, 
but `nosnoc.integrator.SmoothedPss` and `nosnoc.integrator.FESD` are broken. 

I prefer anyway to call it via `nosnoc.Integrator`   but the broken lower level should still be resolved.

nurkanovic:
All integrators should return the step sizes - smoothedPss does not have a getter for 'h'
however, since the matlab integrator return a time grid, with diff(t_grid) one can reocver the h.
this is often useful for nice plots and debugging

```
Error using nosnoc.integrator.SmoothedPss/get (line 142)
nosnoc:h is not a valid field for this integrator.

Error in nosnoc.Integrator/get (line 38)
            ret = obj.plugin.get(field);

Error in oscilator_simulation_example (line 106)
    h_res = integrator.get("h");
 ```

benchmarks/oscilator_integrator_order/integrator_order_experiment.m
apozharski: Currently getting stats only makes sense for the FESD plugin and can be done through `integrator.plugin.stats`. This is now done in this benchmark.

examples/oscilator_simulation/oscilator_simulation_example.m
apozharski: I am not so sure this is correct. The benchmarks folder should be an immutable snapshot in my opinion with each benchmark having a specific commit the will work on and not be touched onwards to reduce our maintenance load but to maintain transparency for any of our publications that rely on that code.

examples/oscilator_simulation/smoothed_pss.m
apozharski: Resolved with latest commit.

examples/oscilator_simulation/oscilator_simulation_example.m
apozharski: Yep, it should no longer be possible to instantiate the lower level plugins directly.

benchmarks/oscilator_integrator_order/integrator_order_experiment.m
apozharski: Changed in examples affected by this.

examples/oscilator_simulation/oscilator_simulation_example.m
nurkanovic: okay, i am fine with this.

nurkanovic (approved):

PR #142, opened 2025-01-13
Add build step to try all examples.

apozharski:
In order to make example maintenance easier this PR adds a non-blocking build step that runs all examples (currently manually updated via text file) , and throws warnings when a file causes an error.

Additionally some other build speedups  via parallelization of tests.


apozharski:
This PR also resolves #143 by defaulting to casadi conic interface for HiGHS, I am pretty sure this is still not shipped directly with casADi but is certainly less of a pain to get.

nurkanovic (approved):
a quite nice feature. just need to move running all examples in some different folder.

PR #145, opened 2025-01-27
[WIP] Integrate `mpecopt` as an option for solving nosnoc MPCCs.

apozharski:
Initial work to integrate `mpecopt`


+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -908,8 +919,141 @@ function create_solver(obj, solver_options, plugin)
...
+                IH = [];
+            end
+            stats = solve@vdx.problems.Mpcc(obj, IG=IG, IH=IH);
+        end
+
+        function [IG,IH,I00] = process_active_set(obj, active_set)
+        % This method takes a nosnoc active set for a PSS and produces an active set for the
+        % complementarity constraints in this problem. It returns these as a boolean array.
+        %
+        % TODO(@anton) there should be a cleaner way of doing this but this is "easy".
```
nurkanovic, line 936: what do you mean

+nosnoc/+activeset/Cls.m
```diff
@@ -0,0 +1,35 @@
+classdef Cls < handle
+    properties(SetAccess=private)
+        active_constraints
```
nurkanovic, line 3: the names of properties in different classes have different style, compare to heaviside, please make consistent.

+nosnoc/+activeset/Heaviside.m
```diff
@@ -0,0 +1,44 @@
+classdef Heaviside < handle
+    properties(SetAccess=private)
+        I_0
+        I_1
+        I_free
+        times
+        stages
```
nurkanovic, line 7: stages of what? rk stages, control stages, other stages? please explain. 
if is a chunk of the horizon where the active set is not changed, we may call it phase instead, coming from multi-phase OCP which is a more common name

examples/michael_schumacher_problem/main_michael_schumacher.m
```diff
@@ -150,9 +153,16 @@
...
+active_set_guess = nosnoc.activeset.Pss({[1,2]},'times', [problem_options.T]);
+solver_options.mpecopt.initialization_strategy = 'TakeProvidedActiveSet';
 %% Solve
 ocp_solver = nosnoc.ocp.Solver(model, problem_options, solver_options);
-ocp_solver.solve();
+
+ocp_solver.set_initial_active_set(active_set_guess);
+[IG,IH,I00] = ocp_solver.discrete_time_problem.process_active_set(active_set_guess);
+
+ocp_solver.solve('mpecopt', IG=IG, IH=IH);
```
nurkanovic, line 165: so this give an error message of too many input arguments, on the other hand, removing this leads to  a warning:

Warning: \n TakeProvidedActiveSet as initialization strategy chosen, but no y0 provided. Resporting to TakeInitialGuessActiveSet 

seems like the passing does not quite work.

nurkanovic: also, there is a typo in the warning message :D

examples/michael_schumacher_problem/main_michael_schumacher.m
```diff
@@ -150,9 +153,16 @@
...
 fprintf('Objective values is: %2.4f \n', ocp_solver.get_objective());
 %% plot
 plot_results_ms(model, problem_options, ocp_solver,...
```
nurkanovic, line 168: a very nice option would be, if the active set method is used, to plot all BNLP iterates -  to see the evolution of active set changes. 

ocp_solver.stats.iter.X_outer 
I believe, but I am not sure how to extract states and controls from them.

+nosnoc/+discrete_time_problem/Stewart.m
apozharski: I apply the active set directly to the algebraic values then get the MPCC active set by evaluating $G$ and $H$. In principle we could  generate the MPCC active set "directly" with some indexing manipulations.

I am less convinced now that this is actually that bad as it is less likely to be broken by some changes to ordering etc. Perhaps I drop the TODO.

+nosnoc/+activeset/Cls.m
apozharski: They are named differently because they _are_ different, in my opinion. I.e. in PDS/CLS it is active unilateral constraints that define the active set, wheras in PSS it is specifically the active regions, and in Heaviside case it is which of the three branches of the Heaviside step each step function is on.

I am happy to change the names but making them similar is wrong in my opinion.

examples/michael_schumacher_problem/main_michael_schumacher.m
apozharski: This can be done by setting `ocp_solver.discrete_time_problem.w.res = ocp_solver_states.iter.X_outer{i}`, assuming they are returned as the primal variables for the unlifted, original form.

I believe we had done this earlier somewhere, and I can quickly implement it again if you wish.

examples/michael_schumacher_problem/main_michael_schumacher.m
apozharski: That warning is in mpecopt :sweat_smile:. This is a bug caused by some rebasing, it has been fixed.

+nosnoc/+activeset/Heaviside.m
```diff
@@ -0,0 +1,44 @@
+classdef Heaviside < handle
+    properties(SetAccess=private)
+        I_0
```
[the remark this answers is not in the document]
apozharski: Index of step functions that are strictly 0.

+nosnoc/+activeset/Heaviside.m
apozharski: Added a comment. I am happy to change the name to phases.

examples/michael_schumacher_problem/main_michael_schumacher.m
```diff
@@ -136,24 +145,191 @@
...
+x_guess = vertcat(xx,yy, zeros(2,problem_options.N_stages), alpha); % Not used ATM
+
+%% Compute active set guess for mpecopt solver
+active_set_guess = nosnoc.activeset.Pss({[1,2]},'times', [problem_options.T]); % [1,2] is a slinding mode (i.e. c = 0), and it is assumed to hold over [0,T]
+solver_options.mpecopt.initialization_strategy = 'TakeProvidedActiveSet';
+
 %% Solve
 ocp_solver = nosnoc.ocp.Solver(model, problem_options, solver_options);
-ocp_solver.solve();
+ocp_solver.set_initial_active_set(active_set_guess);
```
nurkanovic, line 171: what active set is taking if this line is not run? 
the first BNLP Thakes then forever. 
Interestingly, with providing the active set, the optimal solution has normal velocity = 0. i.e., it seems no active set changes. with relax and project there are quite a few switches.

+nosnoc/+activeset/Cls.m
nurkanovic: Okay. I agree.

+nosnoc/+activeset/Heaviside.m
nurkanovic: At this point i still dont understand what are stages :D 
If they are phases, esentially paris of active set and time interval, then we should name them phases. if they are some other sort of stages. should be described in more detail.

+nosnoc/+discrete_time_problem/Cls.m
```diff
@@ -1304,7 +1304,11 @@ function create_solver(obj, solver_options, plugin)
             obj.solver = nosnoc.solver.mpccsol('Mpcc solver', plugin, obj, solver_options);
         end
 
-        function stats = solve(obj)
+        function stats = solve(obj, active_set)
+            arguments
+                obj
```
nurkanovic, line 1309: please add comments what is what

examples/michael_schumacher_problem/main_michael_schumacher.m
apozharski: I don't think it is. the default initialization is the one that is good for Scholtes and co, i,e, _all regions active_.

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -749,12 +749,16 @@ function create_solver(obj, solver_options, plugin, regenerate)
...
+            if regenerate || isempty(obj.solver)
                 obj.solver = nosnoc.solver.mpccsol('Mpcc solver', plugin, obj, solver_options);
             end
         end
 
-        function stats = solve(obj)
+        function stats = solve(obj, active_set)
+            arguments
+                obj
+                active_set {mustBeA(active_set,["nosnoc.activeset.Pds","double"])} = nosnoc.activeset.Pds.empty;
```
nurkanovic, line 760: oh boy, what is this. please add one line comments,

+nosnoc/+discrete_time_problem/Heaviside.m
```diff
@@ -901,7 +901,11 @@ function create_solver(obj, solver_options, plugin)
             obj.solver = nosnoc.solver.mpccsol('Mpcc solver', plugin, obj, solver_options);
         end
 
-        function stats = solve(obj)
+        function stats = solve(obj, active_set)
+            arguments
+                obj
+                active_set {mustBeA(active_set,["nosnoc.activeset.Heaviside","double"])} = nosnoc.activeset.Heaviside.empty;
```
nurkanovic, line 907: same as pds

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -870,21 +871,31 @@ function create_solver(obj, solver_options, plugin)
...
 
-            % Sort by indices to recover almost block-band structure.
-            obj.w.sort_by_index();
-            obj.g.sort_by_index();
+                % Sort by indices to recover almost block-band structure.
+                obj.w.sort_by_index();
+                obj.g.sort_by_index();
 
-            solver_options.assume_lower_bounds = true;
+                solver_options.assume_lower_bounds = true;
```
nurkanovic, line 888: if this is set here, then setting it outside in solver_options has no effect at all?

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: I mean I think it is pretty readable: `mustBeA` does exactly what it seems: it checks if active_set is either an active set or a double (which is necessary to allow passing `[]` because matlab treats empties of different classes differently and has no concept of e.g. `null`) though I can actually make this more readable by making it:

`mustBeA(active_set, ["nosnoc.activeset.Base"])`

and having `nosnoc.activeset.Base.empty` be the default active set.

+nosnoc/+discrete_time_problem/Stewart.m
apozharski: Yes, as it should. For `nosnoc` specific problems this should _always_ be true otherwise the numerics in the relaxed NLP become nasty due to duplicate constraints.

+nosnoc/+discrete_time_problem/Gcs.m
apozharski: Also to explain why you need to use the `mustBeA` verification function rather than passing a type is because I suspect Mathworks only heard of how programming languages are built and decided that type-coersion based on the arguments block is a sane default if you do e.g.:

```MATLAB
function foo = fun(foo)
arguments
   foo SomeClass
end
end
```

and call `x = foo([])` you will _not_ get `x = []` if `SomeClass` has a constructor that takes exactly one argument, you will get `x = SomeClass([])`. Which is _insane_.

+nosnoc/+activeset/Heaviside.m
apozharski: Stages is the name for the alternate approach for defining the end points that you wanted: i.e. instead of times you pass `(control_stage, finite_element)` index pairs.

examples/michael_schumacher_problem/main_michael_schumacher.m
apozharski: Won't be all zeros but the default `w0` which has all regions active.

+nosnoc/+solver/MpccSolver.m
```diff
@@ -1117,6 +1115,7 @@ function generate_cpp_solver(obj, solver_dir)
             import casadi.*;
             p = inputParser;
             addParameter(p, 'x0', []);
+            addParameter(p, 'y0', []);
```
nurkanovic, line 1118: comment what is y0 in the place where it is defined for the first time.

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -908,8 +919,140 @@ function create_solver(obj, solver_options, plugin)
...
+                IH = [];
+            end
+            stats = solve@vdx.problems.Mpcc(obj, IG=IG, IH=IH);
+        end
+
+        function [IG,IH,I00] = process_active_set(obj, active_set)
+        % This method takes a nosnoc active set for a PSS and produces an active set for the
+        % complementarity constraints in this problem. It returns these as a boolean array.
+        %
+        % TODO(@anton) this assumes no user complementarities, but how do we handle those?
```
nurkanovic, line 936: We should maybe throw an error message (if it is easy to implement) as we do not support them.
Otherwise, we will implement it only if some relevant example really requires it.

+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -908,8 +919,140 @@ function create_solver(obj, solver_options, plugin)
                     end
                 end
             end
+            
+            if ~isempty(active_set)
+                [IG,IH,~] = obj.process_active_set(active_set);
+            else
+                IG = [];
```
nurkanovic, line 926: please add  one line comment what happens to the solver if empty empty I_G and I_H are passed.

+nosnoc/+discrete_time_problem/Gcs.m
```diff
@@ -775,8 +779,145 @@ function create_solver(obj, solver_options, plugin, regenerate)
...
+                                done_stage = true;
+                                break
+                            end
+                        end
+                        if done_stage
+                            break
+                        end
+                        kk = 0; % Reset fe counter
+                    end
+                    % handle entering active_constraints TODO(@anton) verify this is correct
```
nurkanovic, line 854: I guess the purpose if this is to account for cross comp of adjecent finite elements. (if yes comment so).

I believe this is correctly implemented, see 
examples/pass_pss_active_set/simulation_pass_active_set.m 
whit nice plots of lambda and theta, everything seems fine.

+nosnoc/+discrete_time_problem/Gcs.m
nurkanovic: Thank you for the explanation. One-two line comment in code is appreciated.

examples/michael_schumacher_problem/main_michael_schumacher.m
nurkanovic: passing a warning when this is done would be nice.

+nosnoc/+activeset/Heaviside.m
nurkanovic: In the pass_pss_active_set example there is an example of use. Please add a comment explaining how it is done.

examples/michael_schumacher_problem/main_michael_schumacher.m
apozharski: Hmmm not sure where this warning should be checked. I guess in solve? will think about it.

PR #147, opened 2025-02-28
[WIP] reorganize the way we handle options in nosnoc for the various MPCC solvers

apozharski:


+nosnoc/+discrete_time_problem/Stewart.m
```diff
@@ -885,7 +885,7 @@ function create_solver(obj, solver_options, plugin)
                 obj.w.sort_by_index();
                 obj.g.sort_by_index();
 
-                solver_options.assume_lower_bounds = true;
+
```
nurkanovic, line 888: we had to have this option in the previous PR  to avoid copying constraints. now it is removed and treated somewhere else? (I had added comments in #145 for this option). please clarify

+nosnoc/+ocp/Solver.m
```diff
@@ -90,11 +90,18 @@
...
-        function solve(obj, plugin)
+        function solve(obj)
             arguments
                 obj
-                plugin = 'reg_homotopy'
             end
+            switch class(obj.solver_opts)
+              case "nosnoc.reg_homotopy.Options"
+                plugin = 'reg_homotopy';
+                obj.solver_opts.assume_lower_bounds = true;
```
nurkanovic, line 100: okay. now I see, the answer to my previous comment. would you be so kind to add here the comment from discrete_time_problem I added in #145

+nosnoc/+discrete_time_problem/Stewart.m
nurkanovic: resolved, see below, and then declare as resolved

nurkanovic:
Everything looks fine.  Conflicts are created, probably by the comment added in #145 which explains why the option of assuming lower bounds must be set to true.

in mpcc_ex1, calling mpecopt directly via mpccsol or its native interface fails. Is this an mpecopt bug that still exists in its main (if yes we need to open an issue there and fix it).  or is it a nosnoc mpccsol bug (dont think so)? 

After resolving this, also this PR can be closed. 🚀

examples/generic_mpcc/mpec_no_s_stationarity.m
```diff
@@ -1,47 +0,0 @@
-% Example 3 from Scheel, Holger, and Stefan Scholtes. "Mathematical programs with complementarity constraints: Stationarity, optimality, and sensitivity." Mathematics of Operations Research 25.1 (2000): 1-22.
```
apozharski, line 1: Was this intentionally deleted?

nurkanovic (approved):
LGTM.

will need to check mpcc_ex1 after update of mpecopt (but needs to update here the used version of mpecopt) -- issue needed?

PR #151, opened 2025-04-02
fix MATLAB 2024b bugs

apozharski:
Mathworks introduce several documented breaking changes and undocumented bugs in 2024b (mostly affecting `vdx`).  This PR works around these at the cost of some loss of QoL features such as autocomplete for vdx variables.


nurkanovic (approved):
LGTM

PR #153, opened 2026-04-21
Fix tests after big MPC merge.

apozharski:
- fix bug in CLS caused by  h_0 re-scale.
- Fix examples list


nurkanovic:
LGTM. Thanks for the great work!

Minor remarks:
- In some example we initialize the CCOpt options but do not crate the nosnoc solver with it.
What do you think about having use_ccopt = true/false in these examples, and then create the corresponding nonsoc solver. Just to make the use more obvious?
- I got confused in our generic mpcc is it `x` or `w`. Apparently it is `x`, but didnt know `w` was breaking the struct. 
- What is the latest allowed Matlab Version, are we up to date?

apozharski:
> In some example we initialize the CCOpt options but do not crate the nosnoc solver with it.
What do you think about having use_ccopt = true/false in these examples, and then create the corresponding nonsoc solver. Just to make the use more obvious?

Done.

Also added 2025a through 2026aa to the test suite. Seems to be running fine.

nurkanovic (approved):

PR #155, opened 2026-04-24
Examples/drones

nurkanovic:
An example of a planar drone carrying a load via a cable.

- The possibility that the cable can stretch or go slack makes this essentially a contact problem and results in nonsmooth dynamics, i.e., a CLS.
- CCOpt support is included in the code but has not been tested extensively. The QPCCs in HyRTI appear to be quite easy to solve.
- FESD-J and Time-freezing with FESD seem to work in the standalone OCP.
- In addition, although not directly relevant to the drone example: 
--reduced verbosity in the Furuta pendulum MPC example and 
--removed some Syscop-internal keywords from the BibTeX entries in the README.


apozharski (approved):
LGTM

